PR #2, opened 2023-06-06
chore(cb2-8084): add integration tests

gjulien-bjss:
## Description

Adding integration tests
Related issue: [CB2-8084](https://dvsa.atlassian.net/browse/CB2-8084)

## Type of change

## Checklist:


src/validators/sysNumTimestamp.ts
```diff
@@ -6,30 +6,30 @@ export const validateSysNumTimestampPathParams = (event: APIGatewayProxyEvent) =
   if (!event.pathParameters?.systemNumber) {
     return addHttpHeaders({
       statusCode: 400,
-      body: 'Missing system number',
+      body: JSON.stringify({ errors: ['Missing system number'] }),
```
gjulien-bjss, line 9: returning valid json here makes it easier to parse, for in integration tests and on the client, further down the line if necessary.

webpack/webpack.development.js
```diff
@@ -10,8 +10,7 @@ module.exports = merge(common, {
   plugins: [
     new CopyPlugin({
       patterns: Object.keys((new AwsSamPlugin({ vscodeDebug: false }).entry())).map((lambdaName) => ([
-        { from: './.env', to: `.aws-sam/build/${lambdaName}/`, noErrorOnMissing: true },
-        { from: './node_modules/@dvsa/cvs-type-definitions/json-schemas/', to: `.aws-sam/build/${lambdaName}/json-schemas`}
+        { from: './.env', to: `.aws-sam/build/${lambdaName}/`, noErrorOnMissing: true }
```
gjulien-bjss, line 13: Looks to me like we don't need to copy the `json-schemas` from `node-modules` if we add `@dvsa/cvs-type-definitions` as a dependency, not a dev-dependency as it it now. Which kind of makes sense since it's actual "code" used at runtime.

Seems to also work for the prod build.

tests/integration/search/get.int.test.ts
```diff
@@ -0,0 +1,19 @@
+describe('get function', () => {
+  describe('happy path', () => {
+    it('should find a record', async () => {
+      const response = await fetch('http:/127.0.0.1:3000/v3/technical-records/XYZEP5JYOMM00020/2019-06-15T10:26:53.903Z');
```
gjulien-bjss, line 4: went with `fetch` since it means we don't actually have to pull in any extra dependencies. Seems to do what we want it to for now anyways. Considered using axios because it does have a slightly nicer API.

naathanbrown (approved):

shivangidas (approved):

PR #6, opened 2023-06-29
feat(cb2-8086): create technical record

Daniel-Searle:
## Create new Tech Record v3 endpoint

Generic post endpoint that calls the test number service in order to generate system numbers.

Create Tech Records endpoint can create a single Technical Record with the details provided

Related issue: [JIRA_TICKET_NUMBER](https://dvsa.atlassian.net/browse/CB2-8086)

## Type of change

## Checklist:


src/services/database.ts
```diff
@@ -93,3 +107,52 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
...
+
+    const response = await lambdaClient.send(command);
+    const bufferResponse = Buffer.from(response.Payload!).toString('utf-8');
+    const bufferBody = JSON.parse(bufferResponse).body;
+    return JSON.parse(bufferBody).systemNumber;
+  } catch (e) {
+    logger.error(`Error in generate system number ${JSON.stringify(e)}`);
+    throw new Error('lambda client failed getting data');
+  }
+};
```
gjulien-bjss, line 158: Probably best to put the number generation logic in a separate file. 
Additionally, we need to be able to generate TrailerIds, ZNumbers and TNumbers. 

If a vehicle is posted without a VRM, we need to set the primaryVrm to:
- For a car, a new ZNumber
- For a HGV, a new ZNumber
- For a LGV, a new ZNumber
- For a motorcyle, a new ZNumber
- For a PSV, a new ZNumber

For trailers, if it does not exist, we need to set the trailerId to:
- For a Small Trailer, a new TNumber
- For a Trailer, a new TrailerId (CNumber)

src/handler/post.ts
```diff
@@ -0,0 +1,34 @@
...
+): Promise<{ body: unknown; statusCode: number }> => {
+  logger.info('Post end point called');
+  try {
+    if (event.body === null) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ error: 'Body is not a valid TechRecord' }),
+      };
+    }
+    // TODO to use proper type when we have them
```
gjulien-bjss, line 17: So, this is where we would like to leverage the npm package we created to runtime-validate the input. IF everything was setup correctly in the npm package, it should be somewhat trivial to validate the body of the request.

gjulien-bjss (commented):
Additionally the linter is not happy

src/services/testNumber.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const response = await lambdaClient.send(command);
+    const bufferResponse = Buffer.from(response.Payload!).toString('utf-8');
+    const bufferBody = JSON.parse(bufferResponse).body;
+    return JSON.parse(bufferBody).systemNumber;
+  } catch (e) {
+    logger.error(`Error in generate system number ${JSON.stringify(e)}`);
+    throw new Error('lambda client failed getting data');
+  }
+};
```
gjulien-bjss, line 38: You should be able to write a generic function that just takes an argument of which number to generate, seeing as the logic is the same in all functions

```suggestion
export enum NumberTypes {
  ZNumber = "z-number",
  TrailerId = "trailerId",
  TNumber = "t-number",
  SystemNumber = "systemNumber"
}

// I wish this was not needed but test number service doesn't return a consistent object 
const apiResponseMap: Record<NumberTypes, string> {
  z-number: "ZNumber",
  trailerId: "trailerId",
  t-number: "TNumber",
  systemNumber: "systemNumber"
}


export const generateNewNumber = async (numberType: NumberTypes): Promise<string> => {
  try {
    if (process.env.AWS_SAM_LOCAL) {
      return '123';
    }
    const input = {
      path: `/${numberType}/`,
      httpMethod: 'POST',
      resource: '/`${numberType}/`,
    };
    const command = new InvokeCommand({
      FunctionName: process.env.TEST_NUMBER_LAMBDA_NAME,
      InvocationType: 'RequestResponse',
      Payload: JSON.stringify(input),
    });
    const response = await lambdaClient.send(command);
    const bufferResponse = Buffer.from(response.Payload!).toString('utf-8');
    const bufferBody = JSON.parse(bufferResponse).body;
    const apiResponseValue = apiResponseMap[numberType]
    return JSON.parse(bufferBody)[apiResponseValue];
  } catch (e) {
    logger.error(`Error in generate ${numberType} ${JSON.stringify(e)}`);
    throw new Error('lambda client failed getting data');
  }
};
```

And yes it would be nice if test number service return a consistent object with just a key of number for instance.
I mean it would actually be really nice if test number service wasn't so specific in its endpoints as well.

src/services/database.ts
```diff
@@ -93,3 +105,51 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
   vin: 'VinIndex',
   trailerId: 'TrailerIdIndex',
 };
+export const postTechRecord = async (request: any) => {
+  const systemNumber = await generateNewNumber(NumberTypes.SystemNumber);
```
naathanbrown, line 109: I would like to extract this logic to the handler level. Ideally I would like the database methods to just do database work. You pass it a record, it saves it, end of work.

naathanbrown: I think also a helper function from the handler would help, to lower code complexity inside that

src/services/database.ts
```diff
@@ -93,3 +105,51 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
...
+    },
+    ExpressionAttributeValues: {
+      ':vin': { S: vin },
+      ':systemNumber': { S: systemNumber },
+    },
+    Item: request,
+  };
+  logger.info('we have got to try and catch');
+  try {
+    const ddbDocClient = DynamoDBDocumentClient.from(ddbClient);
```
naathanbrown, line 142: The bug is in this line, new class needs a new keyword. Ideally tho I would like to not use this and use the non document db client. Should be able to marshall it and use that? Unsure if there is something I am missing here however

naathanbrown:
Overall I think it needs more unit testing for more scenarios as well if possible.

package.json
```diff
@@ -30,7 +30,10 @@
   "license": "MIT",
   "dependencies": {
     "@aws-sdk/client-dynamodb": "^3.341.0",
-    "@aws-sdk/util-dynamodb": "^3.341.0"
+    "@aws-sdk/client-lambda": "^3.362.0",
+    "@aws-sdk/util-dynamodb": "^3.362.0",
+    "eslint-config-airbnb-base": "^15.0.0",
```
naathanbrown, line 35: why airbnb? Is this what we use? Genuine question btw I have no idea. This package is srambiling Snyk tho so need to repair that.

src/handler/post.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+  if (request.techRecord_vehicleType === 'trl' && !request.trailerId && request.techRecord_euVehicleCategory === 'o1') {
+    request.trailerId = await generateNewNumber(NumberTypes.TNumber);
+  }
+  if (request.techRecord_vehicleType === 'trl' && !request.trailerId && request.techRecord_euVehicleCategory === 'o2') {
+    request.trailerId = await generateNewNumber(NumberTypes.TNumber);
+  }
+  if (request.techRecord_vehicleType === 'trl' && !request.trailerId) {
+    request.trailerId = await generateNewNumber(NumberTypes.TrailerId);
+  }
```
naathanbrown, line 55: Could this be moved into it's own file this function? Something like in helpers folder or util folder. I like to keep the handlers as just that.

src/handler/post.ts
```diff
@@ -0,0 +1,62 @@
...
+import { postTechRecord } from '../services/database';
+import logger from '../util/logger';
+import { generateNewNumber, NumberTypes } from '../services/testNumber';
+
+export const handler = async (
+  event: APIGatewayProxyEvent,
+): Promise<{ body: unknown; statusCode: number }> => {
+  logger.info('Post end point called');
+  try {
+    if (!event.body) {
```
naathanbrown, line 17: Do we not need to validate it's a valid body when there too?

src/services/database.ts
```diff
@@ -93,3 +102,28 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
   vin: 'VinIndex',
   trailerId: 'TrailerIdIndex',
 };
+export const postTechRecord = async (request: any) => {
+  if (process.env.AWS_SAM_LOCAL) {
```
naathanbrown, line 106: Why is this needed? I thought we could interact locally anyway?

src/services/database.ts
```diff
@@ -93,3 +102,28 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
...
+      '#vin': 'vin',
+      '#systemNumber': 'systemNumber',
+    },
+    ExpressionAttributeValues: {
+      ':vin': { S: request.vin },
+      ':systemNumber': { S: request.systemNumber },
+    },
+    Item: marshall(request as Record<string, AttributeValue>),
+  };
+  try {
```
naathanbrown, line 122: I think we can now strip this try catch, I placed this in to help debug but I *think* it will now work on the outer parent one. Means we can remove the return null too

package.json
Daniel-Searle: I added this as without it, eslint wasn't working was throwing an error like ESLint couldn't find the config "airbnb-base"

tests/unit/services/testNumber.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    jest.resetModules();
+    process.env = { ...OLD_ENV };
+  });
+
+  afterAll(() => {
+    process.env = OLD_ENV;
+  });
+  describe('Successful response with local environment variable set', () => {
+    it('should return a system number', async () => {
+      process.env.AWS_SAM_LOCAL = 'true';
```
naathanbrown, line 28: These tests aren't actually testing the function right? Because this env var just means it returns on the first conditional rather than actually checking the functionality inside no?

src/handler/post.ts
Daniel-Searle: Do you mean a valid json? Or valid tech record? If you mean valid tech record, how would I identify what that actually looks like for all of the different vehicle types?

src/services/database.ts
Daniel-Searle: Keep the error logger as well? Or would ya like me to remove that?

naathanbrown: Yeah keep the log, just log and throw like the above 2 methods please.

src/services/database.ts
naathanbrown: Yeah get it for the test number service, but I just think the database method should still be able to run? I know Guillaume put a lot of work into making the local dynamo work

src/handler/post.ts
naathanbrown: I meant a valid tech record. We have the types package for this reason as well no? Should be able to import that, work out vehicle type and call the validator (I imagine skeleton here) to confirm that it's valid. Also shouldn't this handler also at some stage be working out record completeness?

package.json
naathanbrown: Cool, just need to mess with the verisioning then to make snyk not hate you 😂

gjulien-bjss: should be a dev dependency though

src/services/database.ts
gjulien-bjss: I agree, I don't see why the writing of the record would have any issues locally. It's able to pull data, don't see why it wouldn't be able to write data. 

You can "mock" out things like testNumber service when running locally, but this is core functionality, so it should work locally.

src/services/testNumber.ts
```diff
@@ -0,0 +1,55 @@
...
+    resource: `/${numberType}`,
+  };
+  const command = new InvokeCommand({
+    FunctionName: process.env.TEST_NUMBER_LAMBDA_NAME,
+    InvocationType: 'RequestResponse',
+    Payload: JSON.stringify(input),
+  });
+  try {
+    const response = await lambdaClient.send(command);
+    const bufferResponse = Buffer.from(response.Payload!).toString('utf-8');
```
gjulien-bjss, line 37:
```suggestion
    const bufferResponse = Buffer.from(response.Payload ?? '').toString('utf-8');
```
makes github happy. And slightly safer

src/services/database.ts
```diff
@@ -93,3 +102,28 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
...
+    return await ddbClient.send(new PutItemCommand(command));
```
gjulien-bjss, line 123:
```suggestion
    return ddbClient.send(new PutItemCommand(command));
```
return await should be avoided (async functions returns a promise anyways, it can be awaited where the function is called)

src/services/database.ts
```diff
@@ -93,3 +102,28 @@ const CriteriaIndexMap: Record<Exclude<SearchCriteria, SearchCriteria.ALL>, Tabl
   vin: 'VinIndex',
   trailerId: 'TrailerIdIndex',
 };
+export const postTechRecord = async (request: any) => {
```
gjulien-bjss, line 105:
```suggestion
export const postTechRecord = async (request: any) => Promise<reponse | null>{
```
Would be nice to have a return type on the function

tests/unit/services/testNumber.test.ts
gjulien-bjss: yeah something like the test below would be good (with the mocked buffer). Unless you're doing this for coverage. in that case I'd suggest something like
```ts
const testCases = [NumberTypes.SystemNumber, NumberTypes,TrailerId, NumberTypes.ZNumer, ...]

it.each(testCases)('should return $number', (number: NumberTypes) => {
  process.env.AWS_SAM_LOCAL = 'true'
  const result = await generateNewNumber(number)
  expect(result).toBe('123')
}
```

The `it.each` probably would be good for the mocked bufferVersion anyways

src/handler/post.ts
Daniel-Searle: Yep, i'll add in the record completeness logic

src/services/database.ts
naathanbrown: One more comment on here, what does this actually return?? Are we returning the record back? Is it just a success? If it is the record back, should this not be unmarshalled?

src/services/database.ts
Daniel-Searle: Are you referring to doing integration tests to test it? So that it will actually hit the docker dynamo?

tests/unit/services/testNumber.test.ts
Daniel-Searle: Never seen that before, that's snazzy! Yes, I was doing it for test coverage. I'll implement something like that above^^

src/services/database.ts
Daniel-Searle: The record doesn't come pack in the payload as standard so it isn't currently being returned. Looking at the spec, you can set ReturnValues to return the "updated record" - happy to return it upstream

src/services/database.ts
gjulien-bjss: I'm just referring to being able to run it locally. Let's say I run the lambda locally, get postman out and I want to check it's saving the record to the local version of the database. That's what I'm talking about

src/services/database.ts
gjulien-bjss: You could just do something like this:
```ts
await ddbClient.send(new PutItemCommand(command))
return request // <- as TechRecord

```

src/services/database.ts
Daniel-Searle: You can run it locally and hit it in postman and it'll be saved in the local dynamoDB?

src/validators/post.ts
```diff
@@ -0,0 +1,13 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+// eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-unsafe-assignment
+
+import { schemas } from '@dvsa/cvs-type-definitions/lib/schemas';
+
+export const identifyObjectType = (obj: any, method: string) => identifySchema(obj.techRecord_vehicleType, obj.techRecord_recordCompleteness, method);
+export const identifySchema = (vehicleType: string, recordCompleteness: string, method: string) => schemas
+  .filter((x: string | string[]) => (x.includes(vehicleType) && x.includes(recordCompleteness)))
+  .filter(((x: string | string[]) => x.includes(method)))
+  .map((x) => x);
```
gjulien-bjss, line 13:
```suggestion
import { schemas } from '@dvsa/cvs-type-definitions/lib/schemas';

export const identifyObjectType = (obj: any, method: string) => identifySchema(obj.techRecord_vehicleType, obj.techRecord_recordCompleteness, method);
export const identifySchema = (vehicleType: string, recordCompleteness: string, method: string) => schemas
  .filter(x => x.includes(vehicleType) && x.includes(recordCompleteness) && x.includes(method))
```
would be good have types/enums for the input of this function

naathanbrown (approved):

tests/unit/services/database.test.ts
```diff
@@ -109,3 +121,32 @@ describe('getBySystemNumberAndCreatedTimestamp', () => {
...
+//         },
+//       }),
+//     };
+//     const mockBuffer = Buffer.from(JSON.stringify(mockBody));
+//     mockLambdaSend.mockReturnValueOnce(Promise.resolve({ Payload: mockBuffer }));
+//     mockGenerateNewNumber.mockResolvedValueOnce('foo');
+//     const res = await postTechRecord(postCarData);
+//     expect(res).toBe('foo');
+//   });
+// });
```
gjulien-bjss, line 152: what's wrong with this test?

src/util/processRequest.ts
gjulien-bjss, line 1: Don't think this file should live inside of a `util` folder, or should be called `processRequest`. 

At the very least should be named something like `processPostRequest` since, you know, we'd be doing a put at some point. Maybe a folder called `processors`? although I don't really like that name. Just it's not a util.

gjulien-bjss (commented):
good stuff, a few comments

package.json
```diff
@@ -30,10 +31,14 @@
   "license": "MIT",
   "dependencies": {
     "@aws-sdk/client-dynamodb": "^3.341.0",
-    "@aws-sdk/util-dynamodb": "^3.341.0"
+    "@aws-sdk/client-lambda": "^3.362.0",
+    "@aws-sdk/util-dynamodb": "^3.362.0",
+    "jwt-decode": "^3.1.2",
+    "ts-node": "^10.9.1"
```
gjulien-bjss, line 37: `ts-node` should be a dev dependency

src/processors/processPostRequest.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+  if ('vin' in request && request.vin) {
+    request.partialVin = request.vin.length < 6 ? request.vin : request.vin.substring(request.vin.length - 6);
+  }
+  if ('techRecord_createdByName' in request) {
+    request.techRecord_createdByName = userDetails.username;
+  }
+  if ('techRecord_createdById' in request) {
+    request.techRecord_createdById = userDetails.msOid;
+  }
```
gjulien-bjss, line 46: I'm confused... why would the property be `in` the request.

I'd say done something like:

```ts
(request as TechrecordGet).systemNumber = systemNumber
(request as TechrecordGet).createdTimeStamp= systemNumber
```

Or you even could give `request` the type of `TechrecordGet` on line 16

tests/unit/services/users.test.ts
```diff
@@ -0,0 +1,13 @@
+import { getUserDetails, UserDetails } from '../../../src/services/user';
+
+describe('Test User Service', () => {
+  describe('Should process user details and return them', () => {
+    it('should successfully process a jwt token and return the relevant fields', () => {
+      const header = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6IkFCQ0RFRiJ9';
```
gjulien-bjss, line 6: I assume from the contents of the test that you generated this token yourself? and it's not a valid token, even if expired

src/processors/processPostRequest.ts
```diff
@@ -0,0 +1,114 @@
+import { isValidObject } from '@dvsa/cvs-type-definitions/schema-validator';
+import logger from '../util/logger';
+import { generateNewNumber, NumberTypes } from '../services/testNumber';
+import { identifySchema } from '../validators/post';
+import { UserDetails } from '../services/user';
+import { HttpMethod, RecordCompleteness, VehicleType } from '../util/enum';
+import { TechrecordGet, TechrecordPut } from '../models/post';
+
+export const processPostRequest = async (input: unknown, userDetails: UserDetails) => {
+  // we should be validating it's a valid technical record HERE.
```
gjulien-bjss, line 10: added a comment. We should be validating it's a valid technical record here. 
then we can use typescript's [type predicate system](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) such as

```ts
const isTechRecord(input): input is TechrecordPut => {
  return isThisObjectValid(object)
}
```

gjulien-bjss: I assume a valid technical record would just mean that it matches the skeleton schema of whichever vehicle we're trying to send.

src/processors/processPostRequest.ts
Daniel-Searle: Some of those values don't seem to be accessible inside of that object in the example of primaryVrm, it's causing an error due to it not being present on a TRL & I don't think we should be doing a mix of both

> Property 'primaryVrm' does not exist on type 'TechrecordGet'.   Property 'primaryVrm' does not exist on type 'GETTRLTechnicalRecordV3Testable'.

tests/unit/services/users.test.ts
Daniel-Searle: Why is it not a valid token?

src/processors/processPostRequest.ts
Daniel-Searle: I'm happy to add that in but based on a previous comment which was on the computeRecordCompleteness method, you mentioned that if it's not a complete or testable record then we can assume that it is a skeleton record. Is it a case of that, that I need to move the computeRecordCompleteness value to be the first? Or do I need to add back the logic to test that the tech record is a skeleton one and if it isn't throw some kind of invalid tech record error?

src/processors/processPostRequest.ts
```diff
@@ -0,0 +1,113 @@
...
+  TechrecordGet,
+  TechrecordHgv,
+  TechrecordMotorcycle,
+  TechrecordPsv,
+  TechrecordPut, TechrecordTrl,
+} from '../models/post';
+
+export const processPostRequest = async (input: unknown, userDetails: UserDetails) => {
+  // we should be validating it's a valid technical record HERE.)
+  if (await isObjectEmpty(input)) {
```
gjulien-bjss, line 22:
```suggestion
  if (isObjectEmpty(input)) {
```

src/processors/processPostRequest.ts
```diff
@@ -0,0 +1,113 @@
...
+import {
+  TechrecordCar,
+  TechrecordGet,
+  TechrecordHgv,
+  TechrecordMotorcycle,
+  TechrecordPsv,
+  TechrecordPut, TechrecordTrl,
+} from '../models/post';
+
+export const processPostRequest = async (input: unknown, userDetails: UserDetails) => {
```
gjulien-bjss, line 20:
```suggestion
export const processPostRequest = async (input: unknown, userDetails: UserDetails): TechRecordGet => {
```

gjulien-bjss (approved):

PR #7, opened 2023-07-17
feat(cb2-8168): amend VIN

owen-corrigan-bjss:
CB2-8168: amend VIN - using patch to technical-record/{systemNumber}/{createdTimestamp}


src/handler/patch.ts
```diff
@@ -0,0 +1,64 @@
...
+import {
+  archiveOldCreateCurrentRecord,
+  getBySystemNumberAndCreatedTimestamp,
+} from '../services/database';
+import { validateUpdateVinRequest, validateVins } from '../validators/patch';
+import { PatchRequestRecords } from '../processors/processPatchVinRequest';
+import { TechrecordGet } from '../models/post';
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  logger.info('Patch Technical Record Called');
```
Daniel-Searle, line 14: Do you think this would be more appropriate to be changed to amend vin called, something along those lines?

src/services/database.ts
```diff
@@ -127,3 +129,34 @@ export const postTechRecord = async (request: TechrecordGet): Promise <Techrecor
...
+          Item: marshall(recordToCreate),
+        },
+      },
+    ],
+  };
+
+  try {
+    await ddbClient.send(new TransactWriteItemsCommand(transactParams));
+
+    return 'records updated';
```
Daniel-Searle, line 156: Does this just update one record? Would it be better to return the entire updated record?

src/processors/processPatchVinRequest.ts
```diff
@@ -0,0 +1,35 @@
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import { getUserDetails } from '../services/user';
+import { TechrecordGet } from '../models/post';
+
+export class PatchRequestRecords {
+  public newRecord: TechrecordGet;
+
```
Daniel-Searle, line 7: Can you remove this space between variables

src/processors/processPatchVinRequest.ts
```diff
@@ -0,0 +1,35 @@
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import { getUserDetails } from '../services/user';
+import { TechrecordGet } from '../models/post';
+
+export class PatchRequestRecords {
```
Daniel-Searle, line 5: Can we move away from using OOP to keep all of these consistent

src/handler/patch.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const isRequestInvalid = validateUpdateVinRequest(event);
+
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+
+  logger.info('Request is Valid');
+
+  const systemNumber: string = decodeURIComponent(
```
Daniel-Searle, line 24: is there a need to decode these?

src/handler/patch.ts
```diff
@@ -0,0 +1,64 @@
...
+    createdTimestamp,
+  ) as TechrecordGet;
+
+  const isVinInvalid = validateVins(currentRecord.vin, newVin.toUpperCase());
+
+  if (isVinInvalid) {
+    return isVinInvalid;
+  }
+
+  logger.info("Vin's have been validated");
```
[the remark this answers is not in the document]
naathanbrown: I actually like this just as another view point, but it should be a logger.debug, so we can toggle it

src/validators/patch.ts
```diff
@@ -0,0 +1,51 @@
...
+    || newVin.length < 3
+    || newVin.length > 21
+    || typeof newVin !== 'string'
+  ) {
+    return addHttpHeaders({
+      statusCode: 400,
+      body: JSON.stringify({ error: 'New VIN is invalid' }),
+    });
+  }
+  if (newVin === oldVin) {
```
naathanbrown, line 42: Genuine question, do we care if a user wants to do this? Seems like if you want to be stupid, you can?

src/services/database.ts
owen-corrigan-bjss: So currently V1 just returns a success message. which is different wording to this one. I'll consult and see what the feeling is for changing this to return the amended records.

template.yml
```diff
@@ -32,9 +32,23 @@ Resources:
...
+        CodeUri: src/handler/
+        Handler: patch.handler
+        Runtime: nodejs18.x
+        Timeout: 20
+        Events:
+          PatchLambda:
+            Type: Api
+            Properties:
+              Path: /v3/technical-records/{systemNumber}/{createdTimestamp}
+              Method: patch
```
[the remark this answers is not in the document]
kcarr-bjss: In V1 it was a PUT however this wasn't exactly right as archiving a technical record shouldn't replace it and requiring the body to contain the entire tech record was unnecessary and not terribly secure as it allowed the client to misuse the endpoint to replace/add any field they liked just by changing the contents of the request. The endpoint now retrieves the technical record and only updates the fields that are necessary to archive the technical record thus making it a PATCH operation rather than a PUT operation.

src/services/database.ts
```diff
@@ -127,3 +129,34 @@ export const postTechRecord = async (request: TechrecordGet): Promise <Techrecor
...
+        Put: {
+          TableName: tableName,
+          Item: marshall(recordToCreate),
+        },
+      },
+    ],
+  };
+
+  try {
+    await ddbClient.send(new TransactWriteItemsCommand(transactParams));
```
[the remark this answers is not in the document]
owen-corrigan-bjss: This is because we need to archive one item as well as create a new one at the same time. TransactWriteItems lets us do both actions and if one fails it leaves the database as it was.

template.yml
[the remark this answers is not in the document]
owen-corrigan-bjss: I believe PUT is going to be used for sending the entire technical record. There is a possibility of this endpoint being expanded to allow users to just send a single field that needs updating so to keep them distinct it was decided to use patch

src/validators/patch.ts
owen-corrigan-bjss: A good question, I mean it would just be a bit of a waste of time, you'd be archiving the current and creating a new record for no real reason? I have no stronger reason than it just feels like a waste of time

src/handler/patch.ts
owen-corrigan-bjss: This was behaviour from the Get endpoint. My gut feeling is no, because I don't think ISO strings should contain unsafe characters and system numbers are I think alpha numeric.

src/handler/patch.ts
```diff
@@ -0,0 +1,67 @@
...
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  logger.info('Amend VIN Called');
+
+  const isRequestInvalid: APIGatewayProxyResult | undefined = validateUpdateVinRequest(event);
+
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+
+  logger.debug('Request is Valid');
```
owen-corrigan-bjss, line 22: Have changed these to logger.debug as per Nathans suggestion so they can be toggled

src/handler/patch.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  logger.debug("Vin's have been validated");
+
+  const patchRecords: Array<TechrecordGet> = processPatchVinRequest(
+    currentRecord,
+    event,
+  );
+
+  const recordToArchive: TechrecordGet = patchRecords[0];
+  const newRecord: TechrecordGet = patchRecords[1];
```
gjulien-bjss, line 48:
```suggestion
  const [recordToArchice, newRecord]= processPatchVinRequest(
    currentRecord,
    event,
  );
```

Daniel-Searle:
I think the name of this file is confusing as there's a couple of other functions that are going to be patches, can this be renamed?

src/handler/patch.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+
+  logger.debug('Request is Valid');
+
+  const systemNumber: string = decodeURIComponent(event.pathParameters?.systemNumber as string);
+  const createdTimestamp: string = decodeURIComponent(event.pathParameters?.createdTimestamp as string);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newVin: string = JSON.parse(event.body!).newVin! as string;
```
naathanbrown, line 28: Can we try some syntax round here to get rid of these non null assertions?

```suggestion
  const newVin: string = JSON.parse(event.body ?? '').newVin ?? '' as string;
```

Or something like that would be even better, or a ternary something like that

src/handler/patch.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  logger.debug("Vin's have been validated");
+
+  const [recordToArchive, newRecord] = processPatchVinRequest(
+    currentRecord,
+    event,
+  );
+
+  try {
+    const patchRequest: string = await archiveOldCreateCurrentRecord(
```
naathanbrown, line 49: If we are not using the value, remove it

src/services/database.ts
naathanbrown, line 156: Is this return actually used for anything? If not should it not just ```return```

src/handler/patch.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const isRequestInvalid: APIGatewayProxyResult | undefined = validateUpdateVinRequest(event);
+
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+
+  logger.debug('Request is Valid');
+
+  const systemNumber: string = decodeURIComponent(event.pathParameters?.systemNumber as string);
```
Daniel-Searle, line 25: I think it would be good to have some validation on these values before they get used, in order to ensure that they're not null/undefined/empty
```suggestion
  const systemNumber: string = event.pathParameters?.systemNumber ?? '';
  const createdTimestamp: string = event.pathParameters?.createdTimestamp ?? '';

  if (!systemNumber || !createdTimestamp) {
    return {
      statusCode: 400,
      body: 'some message',
    };
  }
```

naathanbrown:
> I think the name of this file is confusing as there's a couple of other functions that are going to be patches, can this be renamed?

We may... may... change this. We're unsure, so leave as is for now and if it needs changed in a future PR it can be done

src/handler/patch.ts
Daniel-Searle, line 28: What will happen if newVin is missing? I think some validation is needed to ensure that it is present as it's a key value (see what I did there) to the functionality

src/validators/patch.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  return undefined;
+}
+
+export function validateVins(currentRecord: TechrecordGet, newVin: string) {
+  if (
+    !newVin
+    || newVin.length < 3
+    || newVin.length > 21
+    || typeof newVin !== 'string'
```
Daniel-Searle, line 37: When wouldn't this be a type string? Surely this condition is always true

template.yml
```diff
@@ -32,9 +32,23 @@ Resources:
...
+      Properties:
+        CodeUri: src/handler/
+        Handler: patch.handler
+        Runtime: nodejs18.x
+        Timeout: 20
+        Events:
+          PatchLambda:
+            Type: Api
+            Properties:
+              Path: /v3/technical-records/{systemNumber}/{createdTimestamp}
```
Daniel-Searle, line 53: Is this path going to clash with any of the other patch endpoints?

src/handler/patch.ts
Daniel-Searle, line 49: Why are you returning this value but not doing anything with it?

src/handler/patch.ts
```diff
@@ -0,0 +1,61 @@
...
+      recordToArchive,
+      newRecord,
+    ) as string;
+
+    return addHttpHeaders({
+      statusCode: 200,
+      body: JSON.stringify(formatTechRecord(newRecord)),
+    });
+  } catch (error) {
+    return addHttpHeaders({ statusCode: 500, body: JSON.stringify(error) });
```
Daniel-Searle, line 59: Can you format this like the others above

src/validators/patch.ts
```diff
@@ -0,0 +1,51 @@
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import { addHttpHeaders } from '../util/httpHeaders';
+import { TechrecordGet } from '../models/post';
+import { formatTechRecord } from '../util/formatTechRecord';
+
+export function validateUpdateVinRequest(event: APIGatewayProxyEvent) {
```
Daniel-Searle, line 6:
```suggestion
export const validateUpdateVinRequest = (event: APIGatewayProxyEvent) => {
```

src/validators/patch.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (!newVin) {
+    return addHttpHeaders({
+      statusCode: 400,
+      body: JSON.stringify({ error: 'You must provide a new VIN' }),
+    });
+  }
+  return undefined;
+}
+
+export function validateVins(currentRecord: TechrecordGet, newVin: string) {
```
Daniel-Searle, line 32: export const validateVins = (currentRecord: TechrecordGet, newVin: string) => {

src/validators/patch.ts
Daniel-Searle, line 37:
```suggestion
  if (
    !newVin
    || newVin.length < 3
    || newVin.length > 21
  ) {
```

src/handler/patch.ts
```diff
@@ -0,0 +1,61 @@
...
+  const createdTimestamp: string = decodeURIComponent(event.pathParameters?.createdTimestamp as string);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newVin: string = JSON.parse(event.body!).newVin! as string;
+
+  const currentRecord: TechrecordGet = await getBySystemNumberAndCreatedTimestamp(
+    systemNumber,
+    createdTimestamp,
+  ) as TechrecordGet;
+
+  const isVinInvalid: APIGatewayProxyResult | undefined = validateVins(currentRecord, newVin.toUpperCase());
```
Daniel-Searle, line 35: How are you handling the undefined that might come back from this method?

src/handler/patch.ts
owen-corrigan-bjss: I did think about validating these exist, BUT to trigger the lambda you need to have a systemNumber and a createdTimestamp in the URL so is it necessary?

src/validators/patch.ts
owen-corrigan-bjss: whilst I agree it should always be a string sent from the front end it probably doesn't hurt to check

src/handler/patch.ts
owen-corrigan-bjss: so the value it returns is just a report to say if the transaction has been successful but the function does need to run for the database to be updated

src/handler/patch.ts
owen-corrigan-bjss: If the function is undefined then the handler moves on. If its defined it stops and returns the error. If that makes sense?

naathanbrown: Yeah this is the same as how all the others work, they just set the HTTP headers in the if rather than inside the function

src/services/database.ts
owen-corrigan-bjss: I've set it to return undefined. as the linter doesn't like it when I just return

template.yml
owen-corrigan-bjss: So I believe the idea is this could be potentially extended to patch other items too

shivangidas: Aren't update vin, update vrm and correct the record 3 different endpoints?

src/handler/patch.ts
naathanbrown: I personally like checking it, also to make sure system number is correct. Maybe worth creating a little function that checks these, refactor the other end points that check, and then we just can call that?

owen-corrigan-bjss: checked inside patch validator

src/validators/sysNumTimestamp.ts
```diff
@@ -1,34 +1,35 @@
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+import { addHttpHeaders } from '../util/httpHeaders';
 
 // eslint-disable-next-line consistent-return
-export const validateGetErrors = (event: APIGatewayProxyEvent): APIGatewayProxyResult | undefined => {
+export const validateSysNumTimestampPathParams = (event: APIGatewayProxyEvent) => {
   if (!event.pathParameters?.systemNumber) {
```
owen-corrigan-bjss, line 6: I've renamed this as it feels like there might be more than one path that needs to get systemNumber and Timestamp from the path parameters

kcarr-bjss (approved):

src/validators/patch.ts
```diff
@@ -0,0 +1,58 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newVin: string = JSON.parse(event.body).newVin as string;
+
+  if (!newVin) {
+    return addHttpHeaders({
+      statusCode: 400,
+      body: 'You must provide a new VIN',
+    });
+  }
+  return undefined;
```
gjulien-bjss, line 36:
```suggestion
  return;
```

gjulien-bjss (approved):

Daniel-Searle (approved):

PR #8, opened 2023-07-17
feat(cb2 8167): archive technical record

kcarr-bjss:
## Archive Tech Record v3 endpoint

ArchiveTech Records endpoint can update a single non-archived Technical Record using the provided system number and created timestamp.

Related issue: [CB2-8167](https://dvsa.atlassian.net/browse/CB2-8167)

## Type of change

## Checklist:


Daniel-Searle:
There's lots of listing issues, can they be looked into?

src/handler/archive.ts
```diff
@@ -0,0 +1,70 @@
...
+  const createdTimestamp = decodeURIComponent(event?.pathParameters?.createdTimestamp as string);
+  const userDetails = getUserDetails(event.headers.Authorization as string);
+
+  logger.info(`Get from database with sysNum ${systemNumber} and timestamp ${createdTimestamp}`);
+
+  const record = await getBySystemNumberAndCreatedTimestamp(systemNumber, createdTimestamp) as ArchiveRecord;
+
+  logger.debug(`result is: ${JSON.stringify(record)}`);
+
+  if (!record || !Object.keys(record).length) {
```
Daniel-Searle, line 41: Maybe the method from processPostRequest could be moved into a shared validator as the logic is very similar?

src/handler/archive.ts
```diff
@@ -0,0 +1,70 @@
...
+    return addHttpHeaders({
+      statusCode: 400,
+      body: `No record found matching sysNum ${systemNumber} and timestamp ${createdTimestamp}`,
+    });
+  }
+
+  if (record.techRecord_statusCode === Status.ARCHIVED) {
```
Daniel-Searle, line 48: Are you able to use the type from the type definitions repo? Instead of making a brand new enum

src/handler/archive.ts
```diff
@@ -0,0 +1,70 @@
...
+    return addHttpHeaders({
+      statusCode: 400,
+      body: 'Cannot archive an archived record',
+    });
+  }
+
+  record.techRecord_statusCode = Status.ARCHIVED;
+  record.techRecord_lastUpdatedAt = new Date().toISOString();
```
Daniel-Searle, line 56: Will this cause any issues e.g. when running on the lambda, will it produce the wrong time e.g. producing UTC / GMT?

template.yml
```diff
@@ -38,3 +38,15 @@ Resources:
...
+      Properties:
+        CodeUri: src/handler/
+        Handler: archive.handler
+        Runtime: nodejs18.x
+        Events:
+          PostLambdaApi:
+            Type: Api
+            Properties:
+              Path: /v3/technical-records/archive/{systemNumber}/{createdTimestamp}
+              Method: patch
```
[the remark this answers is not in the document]
kcarr-bjss: This is a patch because the technical record is no longer required as part of the body. In V1 it was and it wasn't great for security because it opened up the possibility of being able to misuse the endpoint as a general purpose tech record replacement. The endpoint now retrieves this from the database and only changes the required fields to archive the record, this makes it a PATCH operation rather than a PUT operation which it really should have been in the first place.

src/handler/archive.ts
```diff
@@ -0,0 +1,70 @@
...
+  const archiveErrors = validateArchiveErrors(event);
+
+  if (archiveErrors) {
+    return addHttpHeaders(archiveErrors);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const body: ArchiveRecordRequestBody = await JSON.parse(event.body as string);
+
+  if (!body.reasonForArchiving) {
```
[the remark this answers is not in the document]
kcarr-bjss: It could be however it would require the body to be parsed twice in order to keep the signature of the validate function to kept clean and in-line with the other validators.

package.json
```diff
@@ -37,10 +37,11 @@
   },
   "devDependencies": {
     "@aws-sdk/types": "^3.341.0",
-    "@dvsa/eslint-config-ts": "^3.0.0",
     "@dvsa/cvs-type-definitions": "^2.0.19",
+    "@dvsa/eslint-config-ts": "^3.0.0",
     "@types/aws-lambda": "^8.10.114",
     "@types/jest": "^28.1.8",
+    "@types/lodash": "^4.14.195",
```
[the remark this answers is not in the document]
kcarr-bjss: This is needed because we need to create deep clones of the imported test data in order to avoid repeating a lot of boiler plate code in each test. If you don't deep clone them and import the test data it causes random failures in the tests depending on the order in which they execute.

src/handler/archive.ts
```diff
@@ -0,0 +1,70 @@
...
+      statusCode: 400,
+      body: 'Reason for archiving not provided',
+    };
+  }
+
+  const systemNumber = decodeURIComponent(event?.pathParameters?.systemNumber as string);
+  const createdTimestamp = decodeURIComponent(event?.pathParameters?.createdTimestamp as string);
+  const userDetails = getUserDetails(event.headers.Authorization as string);
+
+  logger.info(`Get from database with sysNum ${systemNumber} and timestamp ${createdTimestamp}`);
```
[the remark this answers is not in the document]
kcarr-bjss: Could be changed to a debug but it is better to have these logger calls than not, it makes debugging very difficult if there are problems that need to be investigated at a later date.

src/handler/archive.ts
kcarr-bjss: I don't think it would cause an issue as toISOString will produce a timestamp with the necessary time zone adjustments appended to the end so it should be fine.

src/handler/archive.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const archiveErrors = validateArchiveErrors(event);
+
+  if (archiveErrors) {
+    return addHttpHeaders(archiveErrors);
+  }
+
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const body: ArchiveRecordRequestBody = await JSON.parse(event.body as string);
```
Daniel-Searle, line 22: You could do the following to avoid switching off ESLint
```suggestion
    const body: ArchiveRecordRequestBody = await JSON.parse(event.body as string) as ArchiveRecordRequestBody;
```

src/handler/archive.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!body.reasonForArchiving) {
+      return {
+        statusCode: 400,
+        body: 'Reason for archiving not provided',
+      };
+    }
+
+    const systemNumber = event?.pathParameters?.systemNumber as string;
```
Daniel-Searle, line 31:
```suggestion
const systemNumber: string = event?.pathParameters?.systemNumber ?? '';
    const createdTimestamp: string = event?.pathParameters?.createdTimestamp ?? '';
   ```

File: src/handler/archive.ts [hunk 10]
```diff
@@ -0,0 +1,78 @@
...
+    const createdTimestamp = event?.pathParameters?.createdTimestamp as string;
+    const userDetails = getUserDetails(event.headers.Authorization as string);
+
+    logger.debug(`Get from database with sysNum ${systemNumber} and timestamp ${createdTimestamp}`);
+
+    const record = await getBySystemNumberAndCreatedTimestamp(systemNumber, createdTimestamp) as ArchiveRecord;
```
Daniel-Searle, line 37:
```suggestion
    const record: ArchiveRecord = await getBySystemNumberAndCreatedTimestamp(systemNumber, createdTimestamp) as ArchiveRecord;
```

src/handler/archive.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    logger.debug(`result is: ${JSON.stringify(record)}`);
+
+    if (!record || !Object.keys(record).length) {
+      return addHttpHeaders({
+        statusCode: 404,
+        body: `No record found matching sysNum ${systemNumber} and timestamp ${createdTimestamp}`,
```
Daniel-Searle, line 44: I think systemNumber would be more appropriate here instead of sysNum

src/services/user.ts
```diff
@@ -1,11 +1,12 @@
 import jwt_decode from 'jwt-decode';
+import { AuthorisationJwtBearerToken } from '../models/user';
 
 export type UserDetails = { username: string, msOid: string, email: string };
 
 export const getUserDetails = (jwt: string): UserDetails => {
   const removedBearer = jwt.substring(7);
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Daniel-Searle, line 8: remove the disable as you've resolved it

src/handler/archive.ts
naathanbrown, line 22: Not sure a parse needs to be awaited? Think it's not async?

naathanbrown (approved):

src/handler/archive.ts
```diff
@@ -0,0 +1,78 @@
...
+      });
+    }
+
+    if (record.techRecord_statusCode === 'archived') {
+      return addHttpHeaders({
+        statusCode: 400,
+        body: 'Cannot archive an archived record',
+      });
+    }
```
gjulien-bjss, line 53: Similar to the amend vin, could we not just return a `200` if the record is already archived ?

Daniel-Searle: I think logically, it hasn't actually archived anything so I think it might be wrong to return 200 - maybe? I think the 400 response suits this situation where it's tried to do it, but hasn't been successfu

naathanbrown (approved):

Daniel-Searle (approved):

src/validators/archive.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const createdTimestamp = event.pathParameters?.createdTimestamp;
+  if (!isISO8601Date(createdTimestamp)) {
+    return {
+      statusCode: 400,
+      body: 'Invalid created timestamp',
+    };
+  }
+
+  return undefined;
```
gjulien-bjss, line 42:
```suggestion
  return;
```

PR #15, opened 2023-07-25
feat(cb2-8168): Amend VIN

owen-corrigan-bjss:
cb2-8168: adds a check to exclude special characters from VIN's


src/validators/patch.ts
```diff
@@ -38,18 +38,19 @@ export const validateUpdateVinRequest = (event: APIGatewayProxyEvent) => {
 
 // eslint-disable-next-line consistent-return
 export const validateVins = (currentRecord: TechrecordGet, newVin: string) => {
+  console.log((/^[0-9A-Z]+$/).test(newVin));
```
gjulien-bjss, line 41:
```suggestion
```

gjulien-bjss (approved):

PR #20, opened 2023-07-27
feat(cb2-8088): update VRM v3 endpoint

Daniel-Searle:
## Update VRM v3 endpoint

This change is to add a new endpoint that will update VRM

Related issue: [CB2-8088](https://dvsa.atlassian.net/browse/CB2-8088)

## Type of change

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test:ci": "jest --ci",
     "build": "npm run build:prod",
     "build:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.js",
-    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & npm run start:dev",
+    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & DOCKER_HOST=unix://$HOME/.docker/run/docker.sock npm run start:dev",
     "build:prod": "webpack-cli --config webpack/webpack.production.js",
     "start:dev": "npm run dynamo:start && sam local start-api --docker-network dynamo_net",
     "invoke": "sam local invoke --docker-network dynamo_net",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test:ci": "jest --ci",
     "build": "npm run build:prod",
     "build:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.js",
-    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & DOCKER_HOST=unix://$HOME/.docker/run/docker.sock npm run start:dev",
+    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & npm run start:dev",
     "build:prod": "webpack-cli --config webpack/webpack.production.js",
     "start:dev": "npm run dynamo:start && sam local start-api --docker-network dynamo_net",
     "invoke": "sam local invoke --docker-network dynamo_net",
--- a/src/handler/updateVrm.ts
+++ b/src/handler/updateVrm.ts
@@ -1,0 +1,93 @@
+import 'dotenv/config';
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+import logger from '../util/logger';
+import { addHttpHeaders } from '../util/httpHeaders';
+import {
+  updateVehicle,
+  getBySystemNumberAndCreatedTimestamp, searchByAll,
+} from '../services/database';
+import { TechrecordGet } from '../models/post';
+import { formatTechRecord } from '../util/formatTechRecord';
+import { SearchResult } from '../models/search';
+import { processPatchVrmRequest } from '../processors/processVrmRequest';
+import { validateTrailerId, validateUpdateVrmRequest, validateVrm } from '../validators/update';
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  logger.info('Amend VRM Called');
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-assignment
+  const isRequestInvalid: APIGatewayProxyResult | undefined = validateUpdateVrmRequest(event);
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+  logger.debug('Request is Valid');
+
+  const systemNumber: string = decodeURIComponent(event.pathParameters?.systemNumber as string);
+  const createdTimestamp: string = decodeURIComponent(event.pathParameters?.createdTimestamp as string);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newIdentifier: string = JSON.parse(event.body ?? '').newIdentifier as string;
+
+  logger.info('newIdentifier');
+  logger.info(newIdentifier);
+  const currentRecord: TechrecordGet = await getBySystemNumberAndCreatedTimestamp(
+    systemNumber,
+    createdTimestamp,
+  ) as TechrecordGet;
+  let filtered: SearchResult[];
+  const techRecords: SearchResult[] = await searchByAll(newIdentifier);
+  if (currentRecord.techRecord_vehicleType === 'trl') {
+    filtered = techRecords.filter((x) => x.trailerId === newIdentifier);
+    if (filtered.length !== 0) {
+      return addHttpHeaders({
+        statusCode: 400,
+        body: JSON.stringify(`Trailer id ${newIdentifier} already exists`),
+      });
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-call
+    const isVrmInvalid: APIGatewayProxyResult | undefined = validateVrm(currentRecord, newIdentifier.toUpperCase());
+    if (isVrmInvalid) {
+      return isVrmInvalid;
+    }
+  } else {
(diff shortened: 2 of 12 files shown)


gjulien-bjss (commented):
We don't expect to be able to change `trailerId`

src/validators/update.ts
```diff
@@ -50,3 +54,52 @@ export const checkVinValidity = (currentVin: string, newVin: (string | undefined
...
+    return {
+      statusCode: 400,
+      body: 'You must provide a new VRM',
+    };
+  }
+  return false;
+};
+
+// eslint-disable-next-line consistent-return
+export const validateVrm = (currentRecord: TechrecordGet, newVrm: string) => {
```
owen-corrigan-bjss, line 89: might be worth expanding this to check the length and that the VRM contains no special characters? I can't see those checks happening else where

tests/unit/handler/updateVrm.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        systemNumber: 'XYZEP5JYOMM00020',
+        techRecord_vehicleType: 'car',
+        createdTimestamp: '2019-06-24T10:26:54.903Z',
+        techRecord_model: 'null',
+      }]);
+      const result = await handler(request);
+
+      // expect(mockGetBySystemNumberAndCreatedTimestamp).toHaveBeenCalledTimes(1);
+      // expect(mockProcessPatchVrmRequest).toHaveBeenCalledTimes(1);
+      // expect(mockUpdateVehicle).toHaveBeenCalledTimes(1);
```
owen-corrigan-bjss, line 71: are these still needed?

package.json
```diff
@@ -11,7 +11,7 @@
     "test:ci": "jest --ci",
     "build": "npm run build:prod",
     "build:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.js",
-    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & npm run start:dev",
+    "watch:dev": "NODE_ENV=development webpack-cli --config webpack/webpack.development.watch.js & DOCKER_HOST=unix://$HOME/.docker/run/docker.sock npm run start:dev",
```
naathanbrown, line 14: Not sure this can stay in? My docker deamon is not pointed here. Would be interested why this is needed and does the rest of the team need this fix?

src/handler/updateVrm.ts
```diff
@@ -0,0 +1,58 @@
...
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  logger.info('Amend VRM Called');
+  const isRequestInvalid: APIGatewayProxyResult | boolean = validateUpdateVrmRequest(event);
+  if (isRequestInvalid) {
+    return isRequestInvalid;
+  }
+  logger.debug('Request is Valid');
+  const systemNumber: string = decodeURIComponent(event.pathParameters?.systemNumber as string);
+  const createdTimestamp: string = decodeURIComponent(event.pathParameters?.createdTimestamp as string);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
naathanbrown, line 24: Rather than this, can we create a type for the request locally so then we don't have unsafe access? Check the promote endpoint for an example

src/handler/updateVrm.ts
```diff
@@ -0,0 +1,58 @@
...
+  const newVrm: string = JSON.parse(event.body ?? '').newVrm as string;
+  const currentRecord: TechrecordGet = await getBySystemNumberAndCreatedTimestamp(
```
naathanbrown, line 26: shouldn't need to declare a type here, should be implicit as of what the method returns

naathanbrown: Can we update the database method to return a 'TechrecordGet' actually, just realised that hasn't been done

src/handler/updateVrm.ts
```diff
@@ -0,0 +1,58 @@
...
+    systemNumber,
+    createdTimestamp,
+  ) as TechrecordGet;
+  const techRecords: SearchResult[] = await searchByAll(newVrm);
+  logger.info(techRecords);
```
naathanbrown, line 31: Can we give some info alongside a log, just a small sentence or something for context

src/handler/updateVrm.ts
```diff
@@ -0,0 +1,58 @@
...
+  const letFilteredVrm = techRecords.filter((x) => x.primaryVrm === newVrm);
+  if (letFilteredVrm.length !== 0) {
+    return addHttpHeaders({
+      statusCode: 400,
+      body: JSON.stringify(`Primary VRM ${newVrm} already exists`),
+    });
+  }
+  logger.debug('identifier has been validated');
+  try {
```
naathanbrown, line 40: personally I would just wrap the whole end point in the try catch

src/processors/processVrmRequest.ts
```diff
@@ -0,0 +1,27 @@
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import { getUserDetails } from '../services/user';
+import { TechrecordGet } from '../models/post';
+
+export const processPatchVrmRequest = (currentRecord: TechrecordGet, event: APIGatewayProxyEvent): Array<TechrecordGet> => {
+  const userDetails = getUserDetails(event.headers.Authorization!);
```
naathanbrown, line 6: I'd rather ?? '' and then let the method fail than non null assert, same as below

src/processors/processVrmRequest.ts
```diff
@@ -0,0 +1,27 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newVrm: string = JSON.parse(event.body!).newVrm as string;
```
naathanbrown, line 8: Also can we not just pass this in from the handler? Not sure it needs the whole event, so just pass in the parameter

package.json
gjulien-bjss: Agreed, if needed people can export their own env var

src/handler/updateVrm.ts
```diff
@@ -0,0 +1,58 @@
...
+  logger.debug('Request is Valid');
+  const systemNumber: string = decodeURIComponent(event.pathParameters?.systemNumber as string);
+  const createdTimestamp: string = decodeURIComponent(event.pathParameters?.createdTimestamp as string);
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const newVrm: string = JSON.parse(event.body ?? '').newVrm as string;
+  const currentRecord: TechrecordGet = await getBySystemNumberAndCreatedTimestamp(
+    systemNumber,
+    createdTimestamp,
+  ) as TechrecordGet;
+  const techRecords: SearchResult[] = await searchByAll(newVrm);
```
gjulien-bjss, line 30: can we not search by VRM?

tests/unit/handler/updateVrm.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      expect(result.statusCode).toBe(200);
+      expect(result.body).not.toBeNull();
+    });
+  });
+
+  describe('Error handling', () => {
```
naathanbrown, line 77: Can we test when we have a duplicate VRM please?

tests/unit/processors/processVrmRequest.test.ts
```diff
@@ -0,0 +1,51 @@
+import { APIGatewayProxyEvent } from 'aws-lambda';
+import {
+  TechrecordGet,
+} from '../../../src/models/post';
+import { processPatchVrmRequest } from '../../../src/processors/processVrmRequest';
+import postCarData from '../../resources/techRecordCarPost.json';
+
+describe('processVrmRequest', () => {
+  it('returns updated records to archive and to add', () => {
+    const event = {
```
naathanbrown, line 10: My advice would be you don't actually need the whole event for the unit test (even if you don't refactor). I know other people have done this too but you can spoof it easily for a test. look at the get handler tests, just have an object with the needed info, then cast it to unknown then as APIGatewayProxyEvent and it should work. For example all the headers, things like method, resource path, the code does not care they're just data from APIG. Again appreciate others have done this, but I will get down to removing them too.

package.json
Daniel-Searle: @naathanbrown I can't run it via docker without having this set in the package json but I just left it in on accident!

src/processors/processVrmRequest.ts
```diff
@@ -0,0 +1,23 @@
...
+import { TechrecordGet } from '../models/post';
+
+export const processPatchVrmRequest = (currentRecord: TechrecordGet, userDetails: UserDetails, newVrm: string): Array<TechrecordGet> => {
+  const recordToArchive: TechrecordGet = { ...currentRecord };
+  const newRecord: TechrecordGet = { ...currentRecord };
+  const date: string = new Date().toISOString();
+
+  if ('primaryVrm' in newRecord) {
+    newRecord.primaryVrm = newVrm.toUpperCase();
+  }
```
gjulien-bjss, line 11:
```suggestion
    newRecord.primaryVrm = newVrm.toUpperCase();

```

Daniel-Searle: When I tried that I was getting a problem with the type:   Property 'primaryVrm' does not exist on type 'GETTRLTechnicalRecordV3Complete'.

shivangidas: You can use `(newRecord as TechRecordHgv | TechRecordMotorcycle | TechRecordCar | TechRecordPsv).primaryVrm`
or the get schemas specifically

gjulien-bjss (approved):

shivangidas (approved):

owen-corrigan (approved):

naathanbrown (approved):

PR #21, opened 2023-07-28
chore(*): Random improvements to all of v3

naathanbrown:
## Description

- Techrecord -> TechRecord
- StatusCode enum across all the code
- Add and improve validation to the post end point
- Make function for vin validity in it's own file
- Refactor to use audit details

## Type of change

## Checklist:

### DIFF
--- a/src/handler/promote.ts
+++ b/src/handler/promote.ts
@@ -1,12 +1,12 @@
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda/trigger/api-gateway-proxy';
 import 'dotenv/config';
 import { cloneDeep } from 'lodash';
-import { StatusCode } from '../models/StatusCode.enum';
 import { TechRecordGet } from '../models/post';
 import { PromoteRecordRequestBody } from '../models/promote';
 import { SearchCriteria } from '../models/search';
 import { getBySystemNumberAndCreatedTimestamp, searchByCriteria, updateVehicle } from '../services/database';
 import { getUserDetails } from '../services/user';
+import { StatusCode } from '../util/enum';
 import { addHttpHeaders } from '../util/httpHeaders';
 import logger from '../util/logger';
 import { validatePromoteErrors } from '../validators/promote';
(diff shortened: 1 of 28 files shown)


shivangidas (approved):

src/handler/promote.ts
```diff
@@ -51,43 +53,28 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
     if (currentResult.length > 0) {
-      const currentRecord = await getBySystemNumberAndCreatedTimestamp(currentResult[0].systemNumber, currentResult[0].createdTimestamp);
-      currentRecord.techRecord_statusCode = StatusCode.ARCHIVED;
-      currentRecord.techRecord_lastUpdatedAt = new Date().toISOString();
-      currentRecord.techRecord_lastUpdatedByName = userDetails.username;
-      currentRecord.techRecord_lastUpdatedById = userDetails.msOid;
-      recordsToArchive.push(currentRecord);
-
-      logger.debug(`Old current record after update ${JSON.stringify(currentRecord)}`);
+      const currentRecord = await getBySystemNumberAndCreatedTimestamp(currentResult[0].systemNumber, currentResult[0].createdTimestamp) as TechRecordGet;
```
gjulien-bjss: I know you didn't (right now) write this, but can we use `.find()` to find the current technical record (on line 52) instead of `.filter()` ? We're only using `[0]` anyways

gjulien-bjss (commented):
Just a thought, could the setting of audit details not be done in the `createAndArchive` function, of whatever it's called? 

That way whatever handler/processor calls that function doesn't have to worry about setting audit details, it's better encapsulation and separation of concerns.

naathanbrown:
> Just a thought, could the setting of audit details not be done in the `createAndArchive` function, of whatever it's called?
> 
> That way whatever handler/processor calls that function doesn't have to worry about setting audit details, it's better encapsulation and separation of concerns.

Potentially, lets keep it as an idea and come back to it. I like that the database layers just do database work, but also see we are calling that method in a lot of places directly before we save.

gjulien-bjss (approved):

PR #22, opened 2023-07-31
fix(cb2-8819): send validation errors in response

shivangidas:
## Return validation errors for V3 Post and Patch endpoints

- Send a list of validation errors 
- prevent systemNumber and partialVin from updating through request

Related issue: [CB2-8819](https://dvsa.atlassian.net/browse/CB2-8819)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Feature (non-breaking change which adds functionality)

## Checklist:


package.json
```diff
@@ -36,14 +36,15 @@
     "@aws-sdk/lib-dynamodb": "^3.341.0",
     "@aws-sdk/util-dynamodb": "^3.362.0",
     "@types/luxon": "^3.3.0",
+    "ajv": "^8.12.0",
```
gjulien-bjss, line 39: seems slightly annoying to pull in the `ajv` dependency for just some typing... Out of curiousity, does it work to import it as a dev dependency?

src/validators/recordCompleteness.ts
```diff
@@ -28,16 +28,16 @@ export function validateAndComputeRecordCompleteness(input: (TechRecordPut | Tec
 
 const validateSkeletonSchema = (input: (TechRecordPut | TechRecordGet), method: HttpMethod): boolean => {
   const isSkeletonSchema = identifySchema(input.techRecord_vehicleType as VehicleType, RecordCompleteness.SKELETON, method);
-  return isSkeletonSchema ? isValidObject(isSkeletonSchema, input) : false;
+  return isSkeletonSchema ? isValidObject(isSkeletonSchema, input) as boolean : false;
```
gjulien-bjss, line 31: why do you have to cast it as a boolean?
if what you want to check is a make a truthy/falsy value a boolean, use:
```suggestion
  return isSkeletonSchema ? !!isValidObject(isSkeletonSchema, input) : false;
```

shivangidas: isValidObject can return an error object if asked for [https://github.com/dvsa/cvs-type-definitions/blob/develop/schema-validator.ts#L12](https://github.com/dvsa/cvs-type-definitions/blob/develop/schema-validator.ts#L12 ), even though here it will always return boolean, typescript was crying about it

gjulien-bjss: fair enough, might try some typescript magic to make it so we don't have to do that

gjulien-bjss (approved):

PR #36, opened 2023-08-11
fix(cb2-8089): return reason for creation from update vrm cherished transfer

shivangidas:
## Update VTM front end to utilise v3 Tech Records API for light vehicles 

return reason for creation from update vrm cherished transfer
Related issue: [CB2-8089](https://dvsa.atlassian.net/browse/CB2-8089)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/processors/processVrmRequest.ts
+++ b/src/processors/processVrmRequest.ts
@@ -13,8 +13,8 @@ export const processPatchVrmRequest = (currentRecord: TechRecordGet, userDetails
     const updatedNewRecord = setCreatedAuditDetails(newRecord, userDetails.username, userDetails.msOid, new Date().toISOString(), currentRecord.techRecord_statusCode as StatusCode);
     (updatedNewRecord as TechRecordHgv | TechRecordMotorcycle | TechRecordCar | TechRecordPsv).primaryVrm = newVrm.toUpperCase();
     const updatedRecordToArchive = setLastUpdatedAuditDetails(recordToArchive, userDetails.username, userDetails.msOid, new Date().toISOString(), StatusCode.ARCHIVED);
-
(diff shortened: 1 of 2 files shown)


emilyrkelly21 (approved):

src/processors/processVrmRequest.ts
```diff
@@ -13,8 +13,8 @@ export const processPatchVrmRequest = (currentRecord: TechRecordGet, userDetails
     const updatedNewRecord = setCreatedAuditDetails(newRecord, userDetails.username, userDetails.msOid, new Date().toISOString(), currentRecord.techRecord_statusCode as StatusCode);
     (updatedNewRecord as TechRecordHgv | TechRecordMotorcycle | TechRecordCar | TechRecordPsv).primaryVrm = newVrm.toUpperCase();
     const updatedRecordToArchive = setLastUpdatedAuditDetails(recordToArchive, userDetails.username, userDetails.msOid, new Date().toISOString(), StatusCode.ARCHIVED);
-
-    return [updatedRecordToArchive, newRecord];
+    updatedNewRecord.techRecord_reasonForCreation = 'Update VRM - Cherished Transfer';
```
gjulien-bjss, line 16: do we not send a reason for creation from the frontend? If we do, what is the point of asking the user to fill it in if we're just overwriting it here anyways?

gjulien-bjss: Neverming we don't do that

gjulien-bjss (approved):

src/processors/processVrmRequest.ts
shivangidas: Change VRM doesn't have an option to add reason

PR #42, opened 2023-08-22
fix(*): repair axle spacings in array formatter

naathanbrown:
## Description

The axle spacings are an edge case that were not being managed. Add a fix and a test.

## Type of change

## Checklist:


src/util/formatTechRecord.ts
```diff
@@ -5,20 +5,21 @@ const buildArray = (techRecordWithoutArrays: object, arrayName: string, formatte
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let objectToAdd: any = {};
   let arrayIndex = 0;
+  const axleSpacingBias = arrayName.includes('axleSpacing') ? 1 : 0;
   Object.entries(techRecordWithoutArrays).sort().forEach(([key, value]) => {
     if (/_\d+_/.test(key) && key.includes(arrayName)) {
       const splitKey = key.split('_');
```
gjulien-bjss, line 11: You could do something like
```ts
const key = 'techRecord_foo_bar_2_something';

const splitKey = key.split('_');
const idx = splitKey.findIndex(k => !isNaN(parseInt(k)));


```

naathanbrown: I'll have a look at when I'm back. I do think it's worth some consideration of run time but not sure how much difference we're talking really. I do like the simple bias, was thinking just a rename as 'nested array bias' and then depending how nested it is it can work. Thoughts on that?

gjulien-bjss: Considering the operation we're doing here, I don't think running `findIndex` on an array with a maximum length of around 5 or 6 will add significant processing time.

Especially since this saves a call to `includes`, which probably does a loop on the string under the hood anyways

gjulien-bjss: After attending a talk saying benchmarking solved all arguments, I thought I'd benchmark it:
Benchmark2

- Includes is faster but requires hardcoding on the string of axles spacings. Regardless of how you name the variable, you still have to hardcode a string, and it's not very extensible code
- The second fastest is the `+` method
- Slowest is `parseInt` (which we are using in quite a few places in the containing function 😨)

Regardless we're still at 16M op/s second in the `+` method, so I'd argue we're still doing this fast enough.

src/util/formatTechRecord.ts
```diff
@@ -48,11 +51,14 @@ export const formatTechRecord = (techRecordWithoutArrays: object) => {
     if (!/_\d+/.test(key)) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, security/detect-object-injection, @typescript-eslint/no-unsafe-member-access
       formattedTechRecord[key] = value;
+    } else if (/techRecord_\w+_\w+_\d+/.test(key)) {
```
naathanbrown, line 54: Wonder if there is a cool way to do this without if/else now, but not sure the time needed is worth it

shivangidas (approved):

gjulien-bjss (approved):

PR #53, opened 2023-09-08
feat(cb2-8710): improve the efficiency of performing a cherished transfer

owen-corrigan-bjss:
CB2-8710: Improve the efficiency of performing a cherished transfer

Cherished transfer has two scenarios:
 1. where the agent is swapping the VRM for one from another vehicle
 2. where the agent is just changing the VRM on a single vehicle (this is different to correcting an error requires archiving the previous record)
 
 Correcting an error should just be correcting the VRM on a record (this should not archive the record)


src/handler/updateVrm.ts
```diff
@@ -1,69 +1,40 @@
...
+import { processCherishedTransfer } from '../processors/processCherishedTransfer';
+import { processCorrectVrm } from '../processors/processCorrectVrm';
 import {
-  correctVrm, getBySystemNumberAndCreatedTimestamp, searchByCriteria, updateVehicle,
+  getBySystemNumberAndCreatedTimestamp,
 } from '../services/database';
 import { getUserDetails } from '../services/user';
-import { StatusCode } from '../util/enum';
-import { formatTechRecord } from '../util/formatTechRecord';
 import { addHttpHeaders } from '../util/httpHeaders';
```
naathanbrown, line 11: If never used please remove

src/handler/updateVrm.ts
```diff
@@ -1,69 +1,40 @@
...
-      statusCode: 200,
-      body: JSON.stringify(formatTechRecord(newRecord)),
-    });
-  } catch (error) {
-    console.log(error);
-    return addHttpHeaders({
-      statusCode: 500,
-      body: JSON.stringify(error),
-    });
+  if (isCherishedTransfer) {
```
naathanbrown, line 36: I'm not sure I like this new pattern. I'd like to keep the endpoints semantically the same, so we call a processor, that returns the value, we then call to the database, that returns, then we send a response. I don't want calls nested down if possible as I want top layer visibility if that makes sense?

naathanbrown:
Really nice work, please can you clean up the imports that aren't needed and re format that handler a bit and I think this is good to go :)

src/handler/updateVrm.ts
owen-corrigan-bjss: Yeah that's fine, I can change that so that the handler is doing the calls to the db and the processor just formats and returns the records that's no problem

PR #58, opened 2023-09-11
feat(cb2-8710): improve the efficiency of performing a cherished transfer

owen-corrigan-bjss:
CB2-8710: 

Related issue: [CB2-8710]

allows cherished transfers with one or two records separates out logic for correcting a VRM and cherished transfer. 

## Type of change

## Checklist:


src/processors/processCherishedTransfer.ts
```diff
@@ -0,0 +1,126 @@
...
+      donorRecord,
+    );
+  }
+  return cherishedTransferNoDonorVehicle(newVrm, recipientRecord, userDetails);
+};
+
+const formatSecondaryVrms = (record:TechRecordType<'get'>): Array<string> | undefined => {
+  if (record.techRecord_vehicleType !== 'trl') {
+    const secondaryVrms: string[] = record.secondaryVrms ? [...record.secondaryVrms] : [];
+    secondaryVrms.push(record.primaryVrm!);
```
naathanbrown, line 29: I'd rather we ?? ''

naathanbrown:
PR looks good, can you just go over where GH has found issues and tidy those up and then lovely

src/services/database.ts
```diff
@@ -51,7 +51,7 @@ export const searchByCriteria = async (searchCriteria: Exclude<SearchCriteria, S
 
   try {
     const data = await ddbClient.send(new QueryCommand(query));
-    logger.debug(JSON.stringify(data));
+    logger.debug(`DATA FROM DB${JSON.stringify(data)}`);
```
shivangidas, line 54:
```suggestion
    logger.debug(`DATA FROM DB ${JSON.stringify(data)}`);
```

shivangidas (approved):

src/processors/processCherishedTransfer.ts
```diff
@@ -0,0 +1,126 @@
...
+    new Date().toISOString(),
+    recipientRecord.techRecord_statusCode as StatusCode,
+  );
+  if (updatedNewRecord.techRecord_vehicleType !== 'trl') {
+    updatedNewRecord.primaryVrm = newVrm.toUpperCase();
+    updatedNewRecord.secondaryVrms = recipientOldVrms;
+  }
+  updatedNewRecord.techRecord_reasonForCreation = 'Update VRM - Cherished Transfer';
+
+  return [updatedNewRecord, updatedRecordToArchive, {} as TechRecordType<'get'>, {} as TechRecordType<'get'>];
```
gjulien-bjss, line 65: I would definitely avoid using this casting of an empty object to `TechRecordType<"get">` here. This means that we're passing an empty object down to the database service. I would think `dynamoDB` is not too happy if you tell it to `put` two empty objects.

The first thing that comes to mind it to create an interface such as
```ts
interface CherishedTransferRequest {
  recordsToUpdate: TechRecordType<"get">[],
  recordsToArchive: TechRecordType<"get">[],
} 
```
And then you can put the records in their relevant spots, regardless of whether there's a donor vehicle of not, and just pass that to the `updateVehicle` function. TBH you probably don't even need the interface, just an object would work.

gjulien-bjss: this would also probably allow you to bypass having two different functions for with donor or without donor, but instantiating the object first and then pushing to those arrays of `update` or `archive` ad hoc. 

```ts
const request = {recordsToUpdate: [], recordToArchive: []}

if (donorVehicle) {
// do some logic here
request.recordsToUpdate.push(updatedDonorRecord)
request.recordsToArchived.push(archivedDonorRecord)
}

// some more logic
request.recordsToUpdate.push(updatedNewVehicle)
request.recordsToArchived.push(archivedNewVehicle)

return request
```

kpearmanbjss (approved):

src/validators/update.ts
```diff
@@ -94,3 +99,15 @@ export const validateVrm = (currentRecord: TechRecordType<'get'>, newVrm: string
   }
   return false;
 };
+
+export const validateVrmExists = async (vrm: string) => {
+  const techRecords = await searchByCriteria(SearchCriteria.PRIMARYVRM, vrm);
+  const filteredVrm = techRecords.filter((x) => x.primaryVrm === vrm && x.techRecord_statusCode !== StatusCode.ARCHIVED);
+  if (filteredVrm.length) {
```
gjulien-bjss, line 106:
```suggestion
  const vrmExists = techRecords.find((x) => x.primaryVrm === vrm && x.techRecord_statusCode !== StatusCode.ARCHIVED);
  if (vrmExists) {
```

src/handler/updateVrm.ts
```diff
@@ -1,69 +1,87 @@
...
+      body: JSON.stringify(newRecipientRecord),
     });
-  } catch (error) {
-    console.log(error);
+  } catch (err) {
     return addHttpHeaders({
       statusCode: 500,
-      body: JSON.stringify(error),
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      body: JSON.stringify({ error: `Failed to update record : ${err}` }),
```
gjulien-bjss, line 84: could you use the `formatErrorMessage` function to return the error messsage (to keep error messages in a consitent format) ?

PR #59, opened 2023-09-12
feat(cb2-8225): Unarchive Technical Record

kcarr-bjss:
## Description

Adds a new endpoint called unarchive which is capable of resurrecting an archived record where there are no other records which share the same VRM which are not archived.

Related issue: [CB2-8225](https://dvsa.atlassian.net/browse/CB2-8225)

## Type of change

## Checklist:


src/handler/unarchive.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (archivedRecord.techRecord_statusCode !== StatusCode.ARCHIVED) {
+      return addHttpHeaders({
+        statusCode: 400,
+        body: 'Record provided is not an archived record so cannot be unarchived.',
+      });
+    }
+
+    const allVehicleRecords = await searchByCriteria(SearchCriteria.SYSTEM_NUMBER, archivedRecord.systemNumber);
+    const hasUnarchivedRecords: boolean = allVehicleRecords.some((searchResult) => searchResult.techRecord_statusCode !== StatusCode.ARCHIVED
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
naathanbrown, line 52: Can we see if we can clean up the any here? Should be okay now we have proper types

Also, not sure I get why we need to check the vrm?

kcarr-bjss: The requirements state that we have to check for all associated tech records that share the same VRM. The Get type doesn't support the primaryVrm field, though it probably should.

naathanbrown: The get doesn't have primaryVrm!?? That has to be a mistake

Daniel-Searle (approved):

tomcrawleyy (approved):

PR #64, opened 2023-09-21
fix(cb2 8225): Unarchive technical records with different VRMs

kcarr-bjss:
## Description

Fixes an issue where technical records that did not share the same VIM as the tech record being targeted were not being considered when determining whether or not the unarchival is allowed.

Related issue: [CB2-8225](https://dvsa.atlassian.net/browse/CB2-8225)

## Type of change

## Checklist:


shivangidas (approved):

src/handler/unarchive.ts
```diff
@@ -42,7 +42,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     }
 
     const { primaryVrm } = (record as { primaryVrm?: string });
-    const hasUnarchivedRecords = (await searchByCriteria(SearchCriteria.SYSTEM_NUMBER, record.systemNumber))
+    const hasUnarchivedRecords = (await searchByCriteria(SearchCriteria.PRIMARYVRM, primaryVrm as string))
```
naathanbrown, line 45: surely this shifts the problem onto VRM where if it has changed you end up with the same issue? Or am I misunderstanding

Daniel-Searle: Can't you only have one current / provisional record per vehicle tho?

naathanbrown (approved):

PR #93, opened 2024-01-10
feat(cb2-10047): add service for aws SNS

shivangidas:
## Update VRM on test following a cherished transfer

Related issue: [CB2-10047](https://dvsa.atlassian.net/browse/CB2-10047)

## Type of change

## Checklist:


src/services/sns.ts
```diff
@@ -0,0 +1,26 @@
...
+  };
+
+  const command = new PublishCommand(input);
+  try {
+    const response = await snsClient.send(command);
+    logger.debug(response);
+    return undefined;
+  } catch (err: unknown) {
+    logger.error(err);
+    throw new Error(err as string);
```
shivangidas, line 24: any opinions on whether the SNS failure should throw an error or silently fail?

owen-corrigan-bjss: I feel like an error probably makes sense, at least then there is some kind of trail of whats happened

pbardy2000 (approved):

PR #119, opened 2024-03-07
feat(cb2-11176): change plate serial to UUID

naathanbrown:
## Description

Make plate serial number work by using a UUID instead of calling out to the number service.

Related issue: [https://dvsa.atlassian.net/browse/CB2-11176](11176)

## Type of change

## Checklist:

### DIFF
--- a/src/handler/plate.ts
+++ b/src/handler/plate.ts
@@ -1,21 +1,24 @@
+import { HGVAxles } from '@dvsa/cvs-type-definitions/types/v3/tech-record/get/hgv/complete';
 import { TechRecordType as TechRecordTypeByVehicle } from '@dvsa/cvs-type-definitions/types/v3/tech-record/tech-record-vehicle-type';
 import { TechRecordType } from '@dvsa/cvs-type-definitions/types/v3/tech-record/tech-record-verb';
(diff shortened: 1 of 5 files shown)


kcarr-bjss (approved):

src/handler/plate.ts
Daniel-Searle, line 1: I think this handler would benefit from being more modular to make it more readable, and simplified. especially the cannotGeneratePlate function, is feeling overly complicated. for example ,making a validate object kinda thing, containing all of the validation code, create new plates containing the new place + validation kinda thing. I'm also wondering if it would be beneficial to make the function async given that the handler is currently async and probably isn't ideal to use a mix of them. destructing path parameter could be nice too

naathanbrown: Thanks mate, I've met you halfway on a lot of this, take a look please!

So not moving to a new way of validating and generating the object, as this is a pattern employed across the repo. Also sounds more like a class based approach and personally in JS/TS for something of this, a more functional based approach of a lambda too, I don't like that view. Splitting out however is good feedback so I have abstracted more into the validation file, and imported that in.

A mix of async and sync is fine too :)

Also the path destructing is a bit of a issue as you need to call a method on it. so I would just have to reassign the value anyway.

Thanks for the feedback, let me know what you think of what is here and if you are happy please approve.

Daniel-Searle (approved):

PR #124, opened 2024-03-15
feat(cb2-11293): updated spec file and type definitions file

kcarr-bjss:
## Description

Updated the spec file to Open API V3 and updated the type definitions file to the latest version

Related issue: [CB2-11293](https://dvsa.atlassian.net/browse/CB2-11293)

## Type of change

## Checklist:

### DIFF
--- a/docs/spec.yml
+++ b/docs/spec.yml
@@ -4,7 +4,6 @@ info:
     API spec for technical records v3 api
   version: 1.0.0
   title: "Technical Record v3"
-  termsOfService: ''
   contact:
     email: user@email.com
   license:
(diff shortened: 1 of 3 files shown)


naathanbrown (approved):

docs/spec.yml
```diff
@@ -220,24 +219,69 @@ paths:
...
+            type: string
+          required: true
+          description: Date at which the record was created. Sort key of the DynamoDB table.
+      requestBody:
+        required: true
+        description: The tech record to be updated
+        content:
+            application/json:
+              schema:
+                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/develop/json-schemas/v3/tech-record/put/psv/complete/index.json"
```
cb-cs, line 277: Do you need a rebase? This is showing as your change / commit, even though it was merged into develop 14/03.

Also - there appears to be 3 references to /develop in the spec file, rather than /v6.1.0, (this line for example).

Do you happen to know why that is the case?

I'm getting the following error in the IDE when viewing the spec file locally...

`Something went wrong...`
`Invalid JSON pointer: https://raw.githubusercontent.com/dvsa/cvs-type-definitions/develop/json-schemas/v3/tech-record/put/psv/complete/index.json`

kcarr-bjss: Removed the references to develop from the spec file. Attempted a rebase however it would appear the section of code in question was actually deleted in develop due to conflicts with Open API v3.

cb-cs (approved):

PR #130, opened 2024-03-18
DevOps - Update Manifest (88)

cvs-ops:
## Update Manifest.json file in the root of all CVS Functional Repositories


cvs-ops:
Abandoned by Workflow

PR #152, opened 2024-08-23
CB2-13468: review API spec

LGin-BJSS:
## Description

updates the open API spec to reflect the current version of the system

Related issue: [CB2-13468](https://dvsa.atlassian.net/browse/CB2-13468)

## Type of change

## Checklist:

### DIFF
--- a/docs/spec.yml
+++ b/docs/spec.yml
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
 info:
   description: >-
     API spec for technical records v3 api
@@ -46,8 +46,7 @@ paths:
           content:
             application/json:
               schema:
-                type: object
-                $ref: "#/components/schemas/techRecordGetCarComplete"
+                  $ref: '#/components/schemas/techRecordGETComplete'
         '401':
           description: Not authenticated
         '403':
@@ -567,184 +566,15 @@ components:
             read: Grants read access
             write: Grants write access
   schemas:
-    techRecordGetCarComplete:
-      type: object
-      additionalProperties: false
-      required:
-        - systemNumber
-        - vin
-        - createdTimestamp
-        - techRecord_vehicleSubclass
-        - techRecord_vehicleType
-        - techRecord_createdAt
-        - techRecord_createdById
-        - techRecord_createdByName
-        - techRecord_reasonForCreation
-        - techRecord_statusCode
-        - techRecord_noOfAxles
-        - primaryVrm
-      properties:
-        techRecord_applicantDetails_name:
-          nullable: true
-          type: string
-          maxLength: 150
-        techRecord_applicantDetails_address1:
-          nullable: true
-          type: string
-          maxLength: 60
-        techRecord_applicantDetails_address2:
-          nullable: true
-          type: string
-          maxLength: 60
-        techRecord_applicantDetails_postTown:
-          nullable: true
-          type: string
-          maxLength: 60
-        techRecord_applicantDetails_address3:
-          nullable: true
-          type: string
-          maxLength: 60
-        techRecord_applicantDetails_postCode:
-          nullable: true
-          type: string
-          maxLength: 12
-        techRecord_applicantDetails_telephoneNumber:
-          nullable: true
-          type: string
-          maxLength: 25
-        techRecord_applicantDetails_emailAddress:
-          nullable: true
-          type: string
-          maxLength: 255
-        createdTimestamp:
-          type: string
-        partialVin:
-          nullable: true
-          type: string
-        primaryVrm:
-          type: string
-        systemNumber:
-          type: string
-        techRecord_createdAt:
-          type: string
-        techRecord_createdById:
-          type: string
-        techRecord_createdByName:
-          type: string
-        techRecord_euVehicleCategory:
-          nullable: true
-          $ref: "#/components/schemas/euVehicleCategory"
-        techRecord_lastUpdatedAt:
-          nullable: true
-          type: string
-        techRecord_lastUpdatedById:
-          nullable: true
-          type: string
-        techRecord_lastUpdatedByName:
-          nullable: true
-          type: string
-        techRecord_manufactureYear:
-          type: integer
-          nullable: true
-        techRecord_recordCompleteness:
-          type: string
-          enum: 
-            - complete
-        techRecord_noOfAxles:
-          type: integer
-        techRecord_notes:
-          nullable: true
-          type: string
-        techRecord_reasonForCreation:
-          type: string
-        techRecord_regnDate:
-          nullable: true
-          type: string
-        techRecord_statusCode:
-          nullable: true
-          $ref: "#/components/schemas/statusCode"
-        techRecord_vehicleConfiguration:
-          nullable: true
-          $ref: "#/components/schemas/vehicleConfiguration"
-        techRecord_vehicleType:
-          type: string
-          enum: 
-            - car
-        vin:
-          type: string
-        techRecord_vehicleSubclass:
-          title: Vehicle Subclass
-          type: array
-          items:
-            $ref: "#/components/schemas/vehicleSubclass"
-        techRecord_hiddenInVta:
-          nullable: true
-          type: boolean
-        techRecord_updateType:
-          nullable: true
-          type: string          
-        secondaryVrms:
-          nullable: true
-          type: array
-          items:
-            type: string
-    euVehicleCategory:
-      type: string
-      enum:
-        - m1
-        - m2
-        - m3
-        - n1
-        - n2
-        - n3
-        - o1
-        - o2
-        - o3
-        - o4
-        - l1e-a
-        - l1e
-        - l2e
-        - l3e
-        - l4e
-        - l5e
-        - l6e
-        - l7e
-    statusCode:
-      type: string
-      enum:
-        - provisional
-        - current
-        - archived
-    vehicleSubclass:
-      title: Vehicle Subclass
-      type: string
-      enum:
-        - 'n'
-        - p
-        - a
-        - s
-        - c
-        - l
-        - t
-        - e
-        - m
-        - r
-        - w
-    vehicleConfiguration:
-      type: string
-      enum:
-        - rigid
-        - articulated
-        - centre axle drawbar
-        - semi-car transporter
-        - semi-trailer
-        - long semi-trailer
-        - low loader
-        - other
-        - drawbar
-        - four-in-line
-        - dolly
-        - full drawba
+    techRecordGETComplete:
+      description: schemas for complete technical records
+      anyOf:
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/car/complete/index.json"
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/lgv/complete/index.json"
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/motorcycle/complete/index.json"
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/psv/complete/index.json"
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/hgv/complete/index.json"
+        - $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/trl/complete/index.json"
 
 security:
   - OAuth2:
--- a/docs/spec.yml
+++ b/docs/spec.yml
@@ -52,7 +52,7 @@ paths:
         '403':
           description: Unauthorised Access
         '404':
-          description: No resource found which matches identifiers
+          description: No records found matching identifier and criteria 
   '/v3/technical-records/{systemNumber}/{createdTimestamp}':
     get:
       summary: 'Get a complete technical record'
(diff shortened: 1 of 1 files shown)


docs/spec.yml
```diff
@@ -39,14 +39,13 @@ paths:
           description: >-
```
Daniel-Searle, line 40: the enum is primaryVrm not vrm, as well as system number is missing

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+            type: string
+            minLength: 3
+            maxLength: 21
+          required: true
+          description: This represents the systemNumber of the vehicle. Partition key of the DynamoDB table.
+        - in: path
+          name: createdTimestamp
+          schema:
+            type: string
+            pattern: \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{0,20}Z
```
Daniel-Searle, line 345: a valid timestamp is 2019-06-24T10:26:56.903Z. I don't think that, that will match the pattern here

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/psv/complete/index.json"
+        '401':
+          description: Not authenticated
+        '403':
+          description: Unauthorised Access
+        '404':
+          description: Operation not supported
```
Daniel-Searle: A lot of the 404 responses have no resources match the search criteria. Do all of the 404's here, reflect that?

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+            type: string
+            minLength: 3
+            maxLength: 21
+          required: true
+          description: This represents the systemNumber of the vehicle. Partition key of the DynamoDB table.
+        - in: path
+          name: createdTimestamp
+          schema:
+            type: string
+            pattern: \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{0,20}Z
```
Daniel-Searle, line 398: same for this pattern ^^

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+          schema:
+            type: string
+          required: true
+          description: This represents the systemNumber of the vehicle. Partition key of the DynamoDB table.
+        - in: path
+          name: createdTimestamp
+          schema:
+            type: string
+          required: true
+          description: Date at which the record was created. Sort key of the DynamoDB table.
```
Daniel-Searle: For these, i think it would be better to say date time. Otherwise, I think that it implies that it could be 10/10/2024

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/get/psv/complete/index.json"
+        '400':
+          description: Bad request
+        '401':
+          description: Not authenticated
+        '403':
+          description: Unauthorised Access
+        '404':
+          description: Operation not supported
+        '422':
```
Daniel-Searle: when would the 422 response be produced?

docs/spec.yml
```diff
@@ -304,6 +303,260 @@ paths:
...
+          description: Unprocessable request
+        '500':
+          description: Validation Error
```
Daniel-Searle: Validation errors are usually 400 requests

docs/spec.yml
```diff
@@ -144,14 +143,14 @@ paths:
         content:
             application/json:
               schema:
-                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.1.0/json-schemas/v3/tech-record/put/psv/complete/index.json"
+                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.4.0/json-schemas/v3/tech-record/put/psv/complete/index.json"
```
cb-cs: Can you upper case T and R on lines 139 and 515 please, for consistency with the others

`Create Technical Record` 

`Promote a Technical Record`

docs/spec.yml
LGin-BJSS: This is the pattern used in `validateSysNumTimestampPathParams`. Are we seeing issues with requests to this endpoint at the moment with invalid timestamps?

docs/spec.yml
LGin-BJSS: removed

docs/spec.yml
LGin-BJSS: Yep, there are otehr places where validation causes an actual error to be thrown, but not here. Agree that validation errors should be a 400, and we should probably revise the other endpoints

docs/spec.yml
```diff
@@ -77,15 +77,15 @@ paths:
...
         '400':
           description: Bad request
         '401':
           description: Not authenticated
         '403':
           description: Unauthorised Access
         '404':
-          description: Operation not supported
+          description: No resource found which matches identifiers
         '422':
```
Daniel-Searle: Are we sure that a 422 produces that message? I've never seen this server produce this

docs/spec.yml
```diff
@@ -106,29 +106,29 @@ paths:
           schema:
             type: string
           required: true
-          description: Date at which the record was created. Sort key of the DynamoDB table.
+          description: Datetime at which the record was created. Sort key of the DynamoDB table.
```
Daniel-Searle: I think it's worth, specifically calling out that it's a date time string value here

docs/spec.yml
```diff
@@ -106,29 +106,29 @@ paths:
...
+                $ref: "#/components/schemas/techRecordGETComplete"
         '400':
           description: Bad request
         '401':
           description: Not authenticated
         '403':
           description: Unauthorised Access
         '404':
-          description: Operation not supported
+          description: No resource found which matches identifiers
```
Daniel-Searle:
```suggestion
          description: No records found matching identifier and criteria
```

docs/spec.yml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
 info:
   description: >-
     API spec for technical records v3 api
```
Daniel-Searle: I think it might be better to have this say something like, api spec for technical records v3 service

docs/spec.yml
```diff
@@ -33,26 +33,26 @@ paths:
...
         '401':
           description: Not authenticated
         '403':
           description: Unauthorised Access
         '404':
-          description: Operation not supported
+          description: No resource found which matches identifiers
   '/v3/technical-records/{systemNumber}/{createdTimestamp}':
     get:
       summary: 'Get a technical record'
```
Daniel-Searle: I think it's worth specifying that this isn't the sub tech record like the search api, that it is the entire complete technical record

docs/spec.yml
```diff
@@ -77,15 +77,15 @@ paths:
           content:
```
Daniel-Searle, line 77: there's a mix through out of it being called tech record / technical record - can we keep it consistent

docs/spec.yml
```diff
@@ -77,15 +77,15 @@ paths:
...
         '401':
           description: Not authenticated
         '403':
           description: Unauthorised Access
         '404':
-          description: Operation not supported
+          description: No resource found which matches identifiers
         '422':
           description: Unprocessable request
         '500':
```
Daniel-Searle, line 93: 500 for validation errors? This is seen throughout the spec

docs/spec.yml
```diff
@@ -106,29 +106,29 @@ paths:
...
         required: true
         description: The tech record to be updated
         content:
             application/json:
               schema:
-                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.1.0/json-schemas/v3/tech-record/put/psv/complete/index.json"
+                $ref: "#/components/schemas/techRecordPUTComplete"
       responses:
         '200':
           description: Returns the technical record
```
Daniel-Searle: can u specify that it returns the complete technical record pls

docs/spec.yml
```diff
@@ -106,29 +106,29 @@ paths:
...
-          description: Date at which the record was created. Sort key of the DynamoDB table.
+          description: Datetime at which the record was created. Sort key of the DynamoDB table.
       requestBody:
         required: true
         description: The tech record to be updated
         content:
             application/json:
               schema:
-                $ref: "https://raw.githubusercontent.com/dvsa/cvs-type-definitions/v6.1.0/json-schemas/v3/tech-record/put/psv/complete/index.json"
+                $ref: "#/components/schemas/techRecordPUTComplete"
```
Daniel-Searle: Doesn't the patch work like, you can send any number of parameters in the body and it will update them?

docs/spec.yml
Daniel-Searle: Not that i know of, off the top of my head

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+            schema:
+              type: object 
+              required:
+                - newVin
+              properties:
+                newVin:
+                  type: string
+      responses:
+        '200':
+          description: Returns the updated technical record
```
Daniel-Searle: mention, the entire / complete tehcnical record

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+          schema:
+            type: string
+          required: true
+          description: This represents the systemNumber of the vehicle. Partition key of the DynamoDB table.
+        - in: path
+          name: createdTimestamp
+          schema:
+            type: string
+          required: true
+          description: Datetime at which the record was created. Sort key of the DynamoDB table.
```
Daniel-Searle:
```suggestion
          description: Datetime at which the technical record was created. Sort key of the DynamoDB table.
```

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+                    - 'Original'
+                    - 'Manual'
+      responses:
+        '200':
+          description: Success
+        '401':
+          description: Not authenticated
+        '403':
+          description: Unauthorised Access
+        '404':
```
Daniel-Searle, line 479: when would this produce a 404?

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+          description: No resource found which matches identifiers
+  '/v3/technical-records/letter/{systemNumber}/{createdTimestamp}':
+    post:
+      summary: 'Generates a letter for a technical record'
```
Daniel-Searle, line 483: this endpoint is exclusive to trl's isnt it? would be good to call that out specifically

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+                    - 'trailer rejection'
+                paragraphId:
+                  type: number
+                  enum:
+                   - 3
+                   - 4
+                   - 5
+                   - 6
+                   - 7
+      responses:
```
Daniel-Searle, line 529: this can also produce 400 responses

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+          description: Success
+        '401':
+          description: Not authenticated
+        '403':
+          description: Unauthorised Access
+        '404':
+          description: No resource found which matches identifiers
+  '/v3/technical-records/promote/{systemNumber}/{createdTimestamp}':
+    post:
+      summary: 'Promotes the specified technical record to current'
```
Daniel-Searle: worth mentioning that it's promoting provisonal records to current records

docs/spec.yml
```diff
@@ -299,11 +299,261 @@ paths:
...
+            schema:
+              type: object 
+              required:
+                - reasonForPromoting
+              properties:
+                reasonForPromoting:
+                  type: string
+      responses:
+        '200':
+          description: Returns the promoted technical record
```
Daniel-Searle, line 571: complete / subset

docs/spec.yml
```diff
@@ -315,6 +565,25 @@ components:
           scopes:
             read: Grants read access
             write: Grants write access
+  schemas:
+    techRecordGETComplete:
```
Daniel-Searle, line 601: why aren't there any testable and skeleton schemas?

docs/spec.yml
LGin-BJSS: the 422 errors are not something we are acively controlling, but a default code within the HTTP standard. I think the default is unprocessable content.

In practice, this should never happen, as we will be catching the error and handling it in an appropriate way.

LGin-BJSS: Thinking about it 422 is a particularly weird thing to have here on a GET request, as there isnt a content body to BE unprocessable.

docs/spec.yml
LGin-BJSS: In the event of the url paramaters not returning a record
See `validatePlateRecordErrors` in validators/plate.ts

docs/spec.yml
LGin-BJSS: as it can be either, do we need to specify? Seems to make the language more clunky to me

docs/spec.yml
LGin-BJSS: Good shout, will add them

docs/spec.yml
LGin-BJSS: It looks like it works for HGVs as well from the code. Will add reference to the summary though

Daniel-Searle (approved):

docs/spec.yml
```diff
@@ -299,11 +319,272 @@ paths:
...
+                  type: boolean
+                thirdMark:
+                  type: string
+      responses:
+        '200':
+          description: Returns the updated technical record
+          content:
+            application/json:
+              schema:
+                oneof:
```
cb-cs: This is the first one I found, but I think they all need to be `oneOf` rather than `oneof`

When the doc is generated with `oneof` and you click on the schema tab it says `any` rather than listing the schemas.

cb-cs (approved):

PR #155, opened 2024-09-13
feat(CB2-13696): Modify existing batch plate Lambda

Daniel-Searle:
## Modify existing batch plate Lambda

Modify the current **Batch Plate Lambda** to read from the **SQS** rather than direct from **JSON file**

### In this pr I have:

1. Removed the hotfix command from the package json due to it no longer being inside of the hotfix path.
2. Moved the existing batch plate creation code from the hotfix branch to now be in the handler directory.
3. Changed the handler to instead of accepting an array of the batch plate data type, to now accept an SQSEvent and additional logic to then process the SQS record as a result of this.
4. Casting the err variable inside of the catch block as a string, due to without doing that - the error was being masked by an empty object. E.g. Error: {}.
5. Deleted the int tests as this is not longer an endpoint.
6. Deleted the README.md in the hotfix directory.
7. Added unit tests to cover the batch plate creation handler.

Related issue: [CB2-13696](https://dvsa.atlassian.net/browse/CB2-13696)

## Type of change

- [x] Feature (non-breaking change which adds functionality)

## Checklist:


src/handler/batchPlateCreation.ts
```diff
@@ -1,54 +1,48 @@
...
-        const record = await getBySystemNumberAndCreatedTimestamp(systemNumber, createdTimestamp);
-        logger.debug(`result is: ${JSON.stringify(record)}`);
+        const dbRecord = await getBySystemNumberAndCreatedTimestamp(systemNumber, createdTimestamp);
+        logger.debug(`result is: ${JSON.stringify(dbRecord)}`);
 
-        if (!record || !Object.keys(record).length) {
+        if (!dbRecord || !Object.keys(dbRecord).length) {
           logger.error(`Missing record with sysNum ${systemNumber} and timestamp ${createdTimestamp}`);
-          continue;
+          return;
```
LGin-BJSS: we probably want to consider which of of these should log and return, and which need to actually throw errors and get sent to the DL queue.

src/handler/batchPlateCreation.ts
```diff
@@ -78,26 +72,19 @@ export const handler = async (batchPlateRecords: BatchPlateData[]) => {
...
         };
-
         logger.debug(JSON.stringify(plateSqsPayload));
 
         await addToSqs(plateSqsPayload, process.env.DOC_GEN_SQS_QUEUE ?? '');
 
         numberOfSqsAdded += 1;
-      } catch (err) {
-        logger.error(`${systemNumber}, ${createdTimestamp}, ${JSON.stringify(err)}`);
+      } catch (err: unknown) {
```
LGin-BJSS: why are we using unknown and casting the entire object to a string here? can we not catch it as an `Error` and use `err.message`?

Daniel-Searle: During testing, when i left it as it was - it seemed to be producing the logs as error: {}. However, with the casting it seemed to resolve that issue

src/handler/batchPlateCreation.ts
Daniel-Searle: would we want all of these if statements to throw?

LGin-BJSS: Maybe its one for the error handling ticket - But I'd probably say yes. The new DLQ is set up to not retry, I believe, so we'd effectively be throwing these to somewhere that we could see which records were failing

src/handler/batchPlateCreation.ts
m-mullen:
```
catch (error: unknown) {
			if (error instanceof CustomError) {
				this.logger.error(error.message);
			} else if (error instanceof Error) {
				this.logger.error(ErrorEnum.ERROR_OCCURRED, error.message);
			} else {
				this.logger.error(ErrorEnum.UNKNOWN, JSON.stringify(error));
			}
		}
```
this is how it was handled recently in central docs, which does stringify the entire object but only after some type handling. This could be left for logging ticket though

m-mullen (approved):

src/handler/batchPlateCreation.ts
```diff
@@ -0,0 +1,125 @@
...
+  record: T,
+  batchIssuerName: string,
+): boolean => record.techRecord_plates?.some((plate) => plate.plateIssuer === batchIssuerName) ?? false;
+
+/**
+ * This function will update the technical record with the new plate that has been generated.
+ * @param record
+ * @param newPlate
+ */
+const updateRecordWithNewPlate = (record: TechRecordTypeByVehicle<'hgv' | 'trl'>, newPlate: Plates) => {
```
naathanbrown: do these need to be functions? are they going to be re-used?

Daniel-Searle: They're unlikely to be re used, just force of habbit to separate them! 😄 . Happy to put it in on big handler function though!

naathanbrown: Personally I think so. I think if things are complex pulling them out is very useful, but take for example the create plate, all we have done is basically move where the code is on the page.

I also think if they are going to be individual functions we should test them as such, so I would say just place them back in the handler :)

naathanbrown (approved):

PR #156, opened 2024-09-17
feat(cb2-13861): added script for generating batch plate data

LGin-BJSS:
## Description

added a script for generating batch plate data

Related issue: [CB2-13861](https://dvsa.atlassian.net/browse/CB2-13861)

## Type of change

## Checklist:


naathanbrown:
DRAFT DON'T MERGE

tests/util/generateBatchPlateData.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+    else {
+    console.log("written vehicles to seed data")
+}})
+fs.writeFile('tests/resources/trigger-data.json', JSON.stringify(triggerData, null, 2), (err)=> 
+    {if(err){
+        throw err
+    }
+    else {
+    console.log("written vehicles to trigger data")
```
Daniel-Searle: isn't this assuming that these files are already there? If they're not, will they throw an error?

matthew2564: @Daniel-Searle  - As long as the `tests/resources/*` directory exists i.e. the parent, it should create the file gracefully, if not already there.

tests/util/generateBatchPlateData.ts
```diff
@@ -0,0 +1,114 @@
...
+function padZeroes(number: string, size: number) {
+    number = number.toString();
+    while (number.length < size) number = "0" + number;
+    return number;
+}
+
+let records = []
+let triggerData = []
+console.log("starting")
+//complete HGVs
```
Daniel-Searle: Missing the testable and skeleton records

tests/util/generateBatchPlateData.ts
```diff
@@ -0,0 +1,114 @@
...
+for(var i = 1; i<=40000; i++)
```
Daniel-Searle: upon running the script, the first couple of records contain the following details

>   "systemNumber": "XYZEP5JYOMM00020",
>     "createdTimestamp": "2019-06-22T10:26:55.903Z",

Doing this is going to cause them to be rejected when trying to put them into any dynamo db due to the overlap between the keys on the table.

tests/util/generateBatchPlateData.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function padZeroes(number: string, size: number) {
+    number = number.toString();
+    while (number.length < size) number = "0" + number;
+    return number;
+}
+
+let records = []
+let triggerData = []
+console.log("starting")
```
Daniel-Searle: a lot of the functions, don't use console logging but use the winston logger instead

tests/util/generateBatchPlateData.ts
LGin-BJSS: My understanding is that our current seed process only uses the `technical-records-v3.json` file. This script appends to that with the intention of being used through the exisiting seed process. If our seed pulls together multiple files, that should quite quickly solve the issues we were seeing with the size of the JSON.

tests/util/generateBatchPlateData.ts
LGin-BJSS: This is really a throwaway script to show how the test data was generated. These logs are purely for tracking progress when invoking the file through `ts-node` in the CLI.

naathanbrown (approved):

src/handler/uploadPlateSeedData.ts
```diff
@@ -0,0 +1,44 @@
...
+      // eslint-disable-next-line no-await-in-loop
+      await insertBatchPlateSeedData(chunk);
+    }
+  }
+
+  return addHttpHeaders({
+    statusCode: 200,
+    body: 'Uploaded all plate seed data',
+  });
+};
```
Daniel-Searle, line 44: Fix for the Eslint issues, so that we don't need to switch them off! Do we need the event in the parameter - it's not being used anywhere
```suggestion
export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
  logger.info('Upload plate seed end point called');

  const fileNames = generateBatchPlateData();
  await Promise.all(fileNames.map((fileName) => processFile(fileName)));

  return addHttpHeaders({
    statusCode: 200,
    body: 'Uploaded all plate seed data',
  });
};

async function processFile(fileName: string): Promise<void> {
  console.log(fileName);
  const seedData = JSON.parse(fs.readFileSync(fileName, 'utf8')) as Record<string, unknown>[];

  const formattedSeedData = seedData.map((record) => ({
    PutRequest: {
      Item: marshall(record, { removeUndefinedValues: true }),
    },
  }));

  const chunkSize = 25;
  const chunks = [];
  for (let i = 0; i < formattedSeedData.length; i += chunkSize) {
    const chunk = formattedSeedData.slice(i, i + chunkSize);
    chunks.push(chunk);
  }

  await Promise.all(chunks.map(insertBatchPlateSeedData));
}

```

webpack/webpack.production.js
```diff
@@ -10,7 +10,7 @@ const AwsSamPlugin = require("aws-sam-webpack-plugin");
 const LAMBDA_NAMES = ['SearchLambdaFunction', 'GetLambdaFunction', 'PostLambdaFunction', 'PatchLambdaFunction',
  'ArchiveLambdaFunction', 'UnarchiveLambdaFunction', 'PromoteLambdaFunction', 'UpdateVrmFunction',
  'UpdateVinFunction', 'GeneratePlateFunction', 'GenerateLetterFunction', 'SyncTestResultInfoFunction',
- 'GenerateAdrCertificateFunction', 'RemoveInvalidPrimaryVrms', 'BatchPlateCreation', 'MotUpdateVrm'];
+ 'GenerateAdrCertificateFunction', 'RemoveInvalidPrimaryVrms', 'BatchPlateCreation', 'MotUpdateVrm', 'UploadPlateSeed'];
```
Daniel-Searle:
```suggestion
 'GenerateAdrCertificateFunction', 'RemoveInvalidPrimaryVrms', 'BatchPlateCreation', 'MotUpdateVrm', 'UploadPlateSeedDataFunction'];
```

tests/resources/technical-records-v3-with-batch-plates.json
Daniel-Searle, line 1: Now that we're inserting them via the new lambda- do we still need these changes in the jsons?

tests/resources/trigger-data.json
Daniel-Searle, line 1: Appreciate this is being generated locally okay, but when this is running on the lambda - where is the trigger data going to be?

Daniel-Searle:
There are also conflicts on this branch

tests/resources/trigger-data.json
```diff
@@ -0,0 +1,4090 @@
+[
+  {
```
cb-cs: Does it make a difference that these are not camelCase like the original batchPlatesData.json data? To avoid having to manipulate further down the line.

src/util/generateBatchPlateData.ts
```diff
@@ -0,0 +1,222 @@
...
+    const paddedIterator = padZeroes(i.toString(), 6);
+    record.systemNumber = `BPS${paddedIterator}`;
+    record.createdTimestamp = date;
+    record.vin = `BPV${paddedIterator}`;
+    record.partialVin = paddedIterator;
+    record.primaryVrm = `${paddedIterator}Z`;
+    delete record.techRecord_variantNumber;
+
+    records.push(record);
+    triggerData.push({ systemnumber: record.systemNumber, createdtimestamp: record.createdTimestamp });
```
cb-cs: In regards to my comment on trigger-data.json, just wondering if there is a reason this is not systemNumber and createdTimestamp?

tests/resources/trigger-data.json
naathanbrown: into the unknowwnnnnn

tests/resources/technical-records-v3-with-batch-plates.json
naathanbrown: no, will delete them from git before I merge, ty

src/util/generateBatchPlateData.ts
Daniel-Searle: It needs to be camel cased, otherwise it will break the batch load lambda

naathanbrown: caught up with Leo, think it's a boo boo, good spot

src/handler/uploadPlateSeedData.ts
naathanbrown: No question this is better, but I'm just not willing to change so late to SG for something we are going to delete ASAP! Nice work tho man

PR #160, opened 2024-09-25
feat(cb2-13886): repair uri decoding on archive endpoint

naathanbrown:
## Description

This method needs to be on incase a URI ends up encoded which can happen. All other endpoints have it, this has just been missed. Directly fixes the issue Dan Fry and data team are seeing with VOTT.

Related issue: [13886](https://dvsa.atlassian.net/browse/CB2-13886)

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


src/handler/archive.ts
```diff
@@ -24,8 +24,8 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
 
     const body: ArchiveRecordRequestBody = JSON.parse(event.body as string) as ArchiveRecordRequestBody;
 
-    const systemNumber = event?.pathParameters?.systemNumber ?? '';
-    const createdTimestamp = event?.pathParameters?.createdTimestamp ?? '';
+    const systemNumber = decodeURIComponent(event?.pathParameters?.systemNumber ?? '');
+    const createdTimestamp = decodeURIComponent(event?.pathParameters?.createdTimestamp ?? '');
     const userDetails = getUserDetails(event.headers.Authorization ?? '');
```
cb-cs, line 30: Should systemNumber and createdTimestamp always be provided when trying to archive a technical-record, to ensure the correct unique one is identified?

Not necessarily saying it's a change for you to make here, but the existing code implies that they may not be present, so are set to `''`, making the subsequent call to `getBySystemNumberAndCreatedTimestamp` on the database obsolete?

I'll make a note of this and check the previous tickets to see if there was a reason for it.

cb-cs (approved):

cb-cs:
@naathanbrown - code change looks good to me, I know you mentioned in chat you built this off develop, but what has happened with the 19 commits, seems to be including commits from the last few PR's that were merged in?

src/handler/archive.ts
naathanbrown: They are validated in a method just above as being there, but the compiler will not realise it. We opted for this instead of a ! for lining reasons but essentially the ?? path is non-reachable code we are using to keep the compiler happy. Sucks, but if you look, basically every endpoint has this.

cb-cs (approved):

src/handler/archive.ts
cb-cs: Okay fair enough, maybe we can revisit this across the services at some point. Approved.

Daniel-Searle (approved):

PR #162, opened 2024-10-28
feat(cb2-14236): update batch plate failure in handler for sync test result info

cb-cs:
The lambda in the title has a SQS snowball pattern within it. This needs to be refactored to be removed and stop the potential blocking of the lambda with batch failures

Related issue: [CB2-14236](https://dvsa.atlassian.net/browse/CB2-14236)

## Type of change

- [X] Feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have performed a self-review of my own code

### DIFF
--- a/src/handler/sync-test-result-info.ts
+++ b/src/handler/sync-test-result-info.ts
@@ -35,7 +35,9 @@ export const handler = async (event: SQSEvent): Promise<SQSBatchResponse> => {
               test.euVehicleCategory as EUVehicleCategory || undefined,
             );
           } catch (error) {
-            logger.error(`an error occurred while processing test type for record ${record.messageId}: ${JSON.stringify(error)}`);
+            logger.error(`an error occurred while processing record ${record.messageId}: ${error instanceof Error
+              ? error.message
+              : JSON.stringify(error)}`);
             throw new Error('An error occurred while processing test type');
           }
         }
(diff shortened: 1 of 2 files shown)


src/handler/sync-test-result-info.ts
```diff
@@ -34,28 +34,19 @@ export const handler = async (event: SQSEvent): Promise<SQSBatchResponse> => {
               test.createdByName,
               test.euVehicleCategory as EUVehicleCategory || undefined,
             );
-            promisesArray.push(promiseUpdateStatus);
+          } catch (error) {
+            logger.error(`an error occurred while processing record ${record.messageId}: ${error instanceof Error
+              ? error.message
+              : JSON.stringify(error)}`);
+            throw new Error('An error occurred while processing test type');
```
Daniel-Searle: I think it might be more appropriate to remove the nested catch block, i don't see the point in catching it to re throw it to produce a very similar error message in the again in the top level catch block

naathanbrown (approved):

src/handler/sync-test-result-info.ts
cb-cs: Okay fair enough, it was more about having some additional context as to where the code was falling over.

But not precious so happy to remove the nested try/catch and have added the ternary to the top level catch block to ensure we get something tangible in that logger.error

matthew2564 (approved):

naathanbrown (approved):

PR #169, opened 2024-11-19
feat(cb2-14934): create recalls brokering API

naathanbrown:
## Description

Create recalls brokering API. 

Related issue: [14394](https://dvsa.atlassian.net/browse/CB2-14394)

## Type of change

## Checklist:


src/handler/recalls.ts
me-matt, line 1: Can we fix the linting?

naathanbrown: I didn't realise how screwed it was, doing it now

src/util/recalls.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      cache.set('bearerToken', newBearerToken);
+      recallResponse = await fetch(motApiUrl, {
+        headers: {
+          Authorization: `Bearer ${bearerToken}`,
+          'X-API-Key': motSecret.apiKey,
+        },
+      });
+      logger.debug(`second recall response if called: ${JSON.stringify(recallResponse)}`);
+    }
```
matthew2564: I wonder if it makes sense to change what we're caching based on this retry/unsuccessful flow.

By caching the entire `BearerResponse` object, we could avoid a number of bad requests ever hitting the MOT service by using the expiry of the token after decoding.

This would then mean we could determine whether to get a new token, or whether to use the existing one. We could introduce a ~5s buffer tolerance for the refresh.

That means the cache would become something like `const cache: Map<string, BearerResponse> = new Map();`?

What are your thoughts?

naathanbrown: Yeah I thought about this, I decided against it basically as I thought it was overkill. Odds are with a 35 min expire time, that a new instance will probably be created within that time. Also means considering we only have stuff happening between 8-5 we could only ever expire a token 15 times a day anyway. This also covers that if you get a random failure for some reason you get another retry at least once. 

What do you reckon? I could change this over but do you think it's worth it? Happy to do it if you think so

matthew2564: Yeah, I see where you're coming from. I guess I was just trying to stop an extra call out from the lambda (as you could need to hit MOT twice), and maybe reduce the possibility of us hitting MOT with a known expired token.

As you mention this could be very much an edge case. So maybe one for the team to loop back in on for a small discussion.

Do we need to account for the path of say a 500 status from MOT? This code assumes that you'll either get a 401, 403 or a `MotRecalls` shape object. I wonder do we need to check explicitly if 200 then return the object, continue with the 401/403 check, otherwise handle some unexpected status?

src/handler/recalls.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!recalls) {
+      return defaultResponse;
+    }
+
+    const recallsResponse = filterMotRecalls(recalls);
+
+    logger.debug(`Final response: ${JSON.stringify(recallsResponse)}`);
+    return addHttpHeaders({
+      statusCode: 200,
+      body: JSON.stringify(recallsResponse),
```
cb-cs: It makes sense to make use of the Recalls Schema from the `types-definition` package when sending this response back, so it marries up with what the FE are expecting.

```
export interface RecallsSchema {
  hasRecall: boolean;
  manufacturer: string | null;
}
```

src/handler/recalls.ts
```diff
@@ -0,0 +1,89 @@
...
+      return defaultResponse;
+    }
+
+    const vin: string = decodeURIComponent(event.pathParameters?.vin as string);
+    if (!validateSingleVin(vin)) {
+      logger.error(formatErrorMessage('VIN provided in path parameter is not valid.'));
+      return defaultResponse;
+    }
+
+    const motSecret = {
```
cb-cs: I like how the getBearerToken logic sits in the utils, I think the assignment of these motSecret values could probably be delegated to there too, to keep this handler clean.

src/handler/recalls.ts
```diff
@@ -0,0 +1,89 @@
...
+      clientID: process.env.CLIENT_ID,
+      clientSecret: process.env.CLIENT_SECRET,
+      scopeURL: process.env.SCOPE_URL,
+      accessTokenURL: process.env.ACCESS_TOKEN_URL,
+      apiKey: process.env.API_KEY,
+      apiURL: process.env.API_URL,
+    } as MotSecret;
+
+    if (!motSecret) {
```
matthew2564: I'm not sure this code path will ever execute if we are creating the motSecret object above, it will always be truthy.

src/util/recalls.ts
naathanbrown: @cb-cs @m-mullen @me-matt want to offer a view on this either way?

In response to your bottom bit Matt, I think that if you 500 it throws an error, which will be caught? That happened in testing.

cb-cs: I've always been used to tokens lasting ~15 mins, so having 35 mins TTL for the token does change things slightly, and you make a good point as you would think another instance would have been spun up within that time. 

Would it be a lot more work to add it in? I think anything to reduce unnecessary outbound calls is a good thing?

Otherwise, we could always go with this for now, make sure we are logging the errors and can improve further if we do we find we are getting a lot of expired token errors being sent back.

src/util/recalls.ts
```diff
@@ -0,0 +1,116 @@
...
+    logger.error(`Failed to get bearer token: Error: ${(err as Error).message}`);
+    return undefined;
+  }
+};
+
+/**
+ * Create MOT secret
+ * @returns MotSecret - Object with MOT secret details
+ */
+export const populateMotSecret = (): MotSecret => ({
```
cb-cs: Will keep you updated and let you know what comes back from platform, but IMO I think a few of these values should be in a SecretsManager, like clientSecret / apiKey. 

Appreciate there may be reasons it has been done this way, so will add this comment as a reminder to validate.

naathanbrown: secrets? managed.

src/handler/recalls.ts
```diff
@@ -0,0 +1,90 @@
...
+        statusCode: 500,
+        body: 'Recall Feature Flag is undefined',
+      });
+    }
+
+    if (!featureFlags.recallsApi.enabled) {
+      logger.warn('Flag disabled: please enable for recalls functionality');
+      return defaultResponse;
+    }
+
```
Daniel-Searle: Please validate that there is a vin coming in as a path parameter first, before parsing the vin. This should remove the need for the optional chaining and the type casting

src/handler/recalls.ts
```diff
@@ -0,0 +1,90 @@
...
+    const bearerToken = cachedBearerToken ?? await getBearerToken(motSecret);
+
+    if (!bearerToken) {
+      logger.error('bearer token not found');
+      return defaultResponse;
+    }
+
+    if (!cache.has('bearerToken')) {
+      cache.set('bearerToken', bearerToken);
+    }
```
Daniel-Searle: Similarly here, I'd move this into the a separate function to encapsulate the secret logic with the additional benefit of ease of unit testing it

src/handler/recalls.ts
```diff
@@ -0,0 +1,90 @@
...
+      return addHttpHeaders({
+        statusCode: 500,
+        body: 'Recall Feature Flag is undefined',
+      });
+    }
+
+    if (!featureFlags.recallsApi.enabled) {
+      logger.warn('Flag disabled: please enable for recalls functionality');
+      return defaultResponse;
+    }
```
Daniel-Searle: I personally would opt for making the feature flag validation into it's own function to make testing it easier in unit testing so you can test it in isolation

src/models/motRecalls.ts
```diff
@@ -0,0 +1,32 @@
+export interface MotRecalls {
```
Daniel-Searle, line 1: Can see some of the files are called recalls and some of them reference motRecalls. Would be good to keep the naming consistent please to avoid confusion

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+  return {
+    manufacturer: foundRecall ? vehicleRecalls.manufacturer : null,
+    hasRecall: !!foundRecall,
+  };
+};
+
+/**
+ * Retrieve vehicle recall data from MOT recall API
+ * @param vin - vin is query parameter
```
Daniel-Searle: Docblock needs to be updated

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
...
+    return undefined;
+  }
+};
+
+/**
+ * Retrieve bearer token from MOT for recall API
+ * @param vin - vin is query parameter
+ * @returns Promise<BearerToken> - JWT bearer token for recalls
+ */
+export const getBearerToken = async (motSecret: MotSecret): Promise<string | undefined> => {
```
Daniel-Searle: docblock needs updating

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
...
+  logger.debug('Calling MOT Token');
```
Daniel-Searle:
```suggestion
  logger.debug('Calling MOT Token Retrieval');
```

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
...
+      });
+
+      parsedRecallResponse = await recallResponse.json() as MotRecalls;
+
+      logger.debug(`second recall status code: ${recallResponse.status}`);
+      logger.debug(`second recall response: ${JSON.stringify(parsedRecallResponse)}`);
+    }
+
+    if (recallResponse.status === 200) {
+      return parsedRecallResponse;
```
Daniel-Searle: are we not going to be adding any validation to the response?

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Search retrieved recall data for an active recall then construct return object.
+ * @param vehicleRecalls
+ * @returns
+ */
+export const filterMotRecalls = (vehicleRecalls: MotRecalls): RecallsSchema => {
+  logger.debug('Filter Recall Response');
+  const foundRecall = vehicleRecalls.recalls.find((recall) => {
+    if (recall.repairStatus === 'NOT_FIXED' && Date.parse(recall.recallCampaignStartDate) < Date.now()) {
```
Daniel-Searle: I'd suggest using the Luxton library and specifying the timezone so it's quickly identifiable
```suggestion
    if (recall.repairStatus === 'NOT_FIXED' && DateTime.fromISO(recall.recallCampaignStartDate) < DateTime.now()) {
```

src/validators/recalls.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * validate the input vin has a valid format
+ * @param vin - query param
+ * @returns boolean - valid/invalid format
+ */
+export const validateSingleVin = (vin: string) => {
+  if (vin) {
+    if (vin.length < 3
+      || vin.length > 21
+      || typeof vin !== 'string'
```
Daniel-Searle: The parameter for this function is a string, so this will always be false

src/validators/recalls.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * validate the input vin has a valid format
+ * @param vin - query param
+ * @returns boolean - valid/invalid format
+ */
+export const validateSingleVin = (vin: string) => {
+  if (vin) {
+    if (vin.length < 3
```
Daniel-Searle: Does this align with the vin validation that are in the other services and clients?

tests/unit/handler/recalls.unit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  const mockDefaultResponse: APIGatewayProxyResult = addHttpHeaders({
+    statusCode: 200,
+    body: JSON.stringify({
+      manufacturer: null,
+      hasRecall: false,
+    }),
+  });
+
+  describe('WHEN the feature flag is not defined', () => {
```
Daniel-Searle: The feature flag tests should be grouped using it's own describe. you could also use the test.ech functionality isnce these tests are very similar to eachother

tests/unit/handler/recalls.unit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        },
+      });
+
+      const res = await handler({} as APIGatewayProxyEvent);
+      expect(res.statusCode).toEqual(mockDefaultResponse.statusCode);
+      expect(res.body).toEqual(mockDefaultResponse.body);
+    });
+  });
+
+  describe('WHEN the VIN validation fails', () => {
```
Daniel-Searle: group these in their own describe block

tests/unit/handler/recalls.unit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    it('SHOULD log VIN ERROR and return a 200 response with no recalls', async () => {
+      mockGetProfile.mockResolvedValue({
+        recallsApi: {
+          enabled: true,
+        },
+      });
```
Daniel-Searle: once grouped, you can have a before each to add this mock into - it's the same in each test

tests/unit/util/recalls.unit.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      manufacturer: null,
+      hasRecall: false,
+    };
+
+    beforeEach(() => {
+      date = new Date();
+    });
+
+    it('should correctly filter a list of 3 when one is a recall', () => {
+      const recallResponse = {
```
Daniel-Searle: I'd suggest having a helper function to construct these responses and recall objects, it really bloats out the tests having them like this

tests/unit/validators/recalls.unit.test.ts
```diff
@@ -0,0 +1,54 @@
+import { validateSingleVin } from '../../../src/validators/recalls';
+
+describe('validateSingleVin', () => {
```
Daniel-Searle: use test.each here, these unit tests are all very similar. you can refine the validation rules at the top of the first describe eg INVALID_CHARS=[Q] etc

tests/unit/handler/recalls.unit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  const mockDefaultResponse: APIGatewayProxyResult = addHttpHeaders({
+    statusCode: 200,
+    body: JSON.stringify({
+      manufacturer: null,
+      hasRecall: false,
+    }),
+  });
+
+  describe('WHEN the feature flag is not defined', () => {
+    it('SHOULD return a 500 response', async () => {
```
cb-cs: Can you write a test case for when an error is caught in the catch block at the top level of the handler as well please.

Lines 83 - 89

```
  } catch (err) {
    console.error(err);
    return addHttpHeaders({
      statusCode: 500,
      body: 'Error calling recalls API',
    });
  }
```

tests/unit/util/recalls.unit.test.ts
```diff
@@ -0,0 +1,238 @@
...
+        json: () => ({
+          access_token: 'test',
+        }),
+      } as unknown as Response));
+
+      const res = await RecallUtils.getBearerToken({} as MotSecret);
+      expect(res).toBe('test');
+    });
+  });
+  describe('getMotRecallsByVin', () => {
```
cb-cs: Can you add testcase to test what is returned from the catch block of this method please.

Lines 76-77
```
  } catch (err) {
    logger.error(`failed calling MOT endpoint: Error: ${(err as Error).message}`);
    return undefined;
  }
```

src/util/recalls.ts
```diff
@@ -0,0 +1,110 @@
+import { RecallsSchema } from '@dvsa/cvs-type-definitions/types/v1/recalls';
```
cb-cs: Thinking about this, and I am not precious about it but in keeping with the structure of the existing repo...

Arguably the `getMotRecallsByVin` and `getBearerToken` methods could reside in some sort of `api` service (in the service directory), with a `processRecalls` (in the processors directory) subsequently utilising the `filterMotRecalls` method in utils.

It would clean up the structure a bit.

cb-cs: Also appreciate we have to balance more work vs getting this done.

src/handler/recalls.ts
naathanbrown: Yes but it is not possible to call this endpoint without having a path param there, or APIG would not invoke. I would rather the optional chain than code that is not needed.

src/util/recalls.ts
naathanbrown: What do you suggest? If 200 and they honour the contract, what would we want to validate for? If it's anything else this returns undefined and we say no recalls to our clients

src/util/recalls.ts
naathanbrown: Disagree, best to make use of just JS where we can. We are in the lucky position where timezones don't usually matter for us so what we have will be okay

src/handler/recalls.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (!bearerToken) {
+      logger.error('bearer token not found');
+      return defaultResponse;
+    }
+
+    if (!cache.has('bearerToken')) {
+      cache.set('bearerToken', bearerToken);
+    }
+
+    const recalls = await getMotRecallsByVin(vin, cache, motSecret);
```
cb-cs: I like what you have done with the caching above, where did we land on checking the expiryDate of the token prior to using it to make this getMotRecallsByVin call?

It is something we are going to want to put in at some point, wondering if it's worth adding the logic in for that now we have this wired up and working?

src/validators/recalls.ts
naathanbrown: yes

src/handler/recalls.ts
naathanbrown: I disagree. Maybe to better test the caching but I am not really too fussed about that being managed at the top level where it is declared. By abstracting this into a method it would just be moving a module function call down a level and I see no value in that.

tests/unit/handler/recalls.unit.test.ts
naathanbrown: group with what sorry?

tests/unit/util/recalls.unit.test.ts
naathanbrown: I considered this, but I vary a lot of the data within them. The function in my head would be too complex to justify creating for 5 tests.

tests/unit/handler/recalls.unit.test.ts
naathanbrown: Now not needed due to refactor

src/util/recalls.ts
naathanbrown: The processors are used liberally and not consistently anyway. They more cover full processes. such as swapping a vin, or VRM. This would be overkill and not needing one imo.

src/handler/recalls.ts
naathanbrown: Imo we leave it. 35 min tokens makes this a waste of time and something that will basically never happen.

Daniel-Searle (approved):

src/handler/recalls.ts
cb-cs: Okay, there is sufficient logging there for us to monitor this for now. If we do start seeing the error message from a 401/403's being returned when the FE start using this, then it's definitely something we will need to look at.

cb-cs (approved):

m-coslett (approved):

PR #199, opened 2025-10-27
feat(cb2-19474): added logic for skip automated processes

cb-cs:
## Description

Related issue: [CB2-19474](https://dvsa.atlassian.net/browse/CB2-19474)

## Type of change

- [X] Feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have added tests that prove my fix is effective or that my feature works

[CB2-19474]: https://dvsa.atlassian.net/browse/CB2-19474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/handler/sync-test-result-info.ts
+++ b/src/handler/sync-test-result-info.ts
@@ -1,10 +1,13 @@
 import { EUVehicleCategory } from '@dvsa/cvs-type-definitions/types/v3/tech-record/enums/euVehicleCategory.enum.js';
 import { SQSBatchResponse, SQSEvent } from 'aws-lambda';
 import 'dotenv/config';
+import { FeatureFlags, getProfile } from '@dvsa/cvs-feature-flags/profiles/vtx';
 import { TestResult } from '../models/testResult';
 import { processRecord } from '../processors/processSQSRecord';
 import { syncTestResultInfo } from '../processors/processSyncTestResultInfo';
 import logger from '../util/logger';
+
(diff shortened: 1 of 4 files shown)


matthew2564 (approved):

src/handler/sync-test-result-info.ts
```diff
@@ -1,18 +1,33 @@
 import { EUVehicleCategory } from '@dvsa/cvs-type-definitions/types/v3/tech-record/enums/euVehicleCategory.enum.js';
 import { SQSBatchResponse, SQSEvent } from 'aws-lambda';
 import 'dotenv/config';
+import { FeatureFlags, getProfile } from '@dvsa/cvs-feature-flags/profiles/vtx';
 import { TestResult } from '../models/testResult';
 import { processRecord } from '../processors/processSQSRecord';
 import { syncTestResultInfo } from '../processors/processSyncTestResultInfo';
 import logger from '../util/logger';
 
+let featureFlags: FeatureFlags;
```
m-coslett, line 10: should this be static?

cb-cs: This is not in a class, but being a module scoped variable it should persist and behave the same way as a static variable would in a class.

m-coslett (approved):

